PR #83, opened 2025-10-08
feat(bridge): replace ?tab=tx_history with /transaction-history

fionnachan:
Closes FS-1477


packages/app/src/app/(with-sidebar)/bridge/tx-history/page.tsx
```diff
@@ -0,0 +1,14 @@
...
+import BridgePageWrapper from '../BridgePageWrapper';
+import { generateMetadata } from '../page';
+
+type Props = {
+  searchParams: { [key: string]: string | string[] | undefined };
+};
+
+export { generateMetadata };
+
+export default async function BridgeBuyPage({ searchParams }: Props) {
```
chrstph-dvx: does this need to be async?

fionnachan: ah this was a leftover from my WIP months ago. updated.

packages/app/src/app/(with-sidebar)/bridge/buy/[slug]/page.tsx
```diff
@@ -20,10 +18,10 @@ export default async function BridgeBuyOnrampServicePage({ searchParams, params
...
-    <BridgePageWrapper
-      searchParams={resolvedSearchParams}
-      redirectPath={`${PathnameEnum.BUY}/${slug}`}
-    />
-  );
+  await initializeBridgePage({
+    searchParams: resolvedSearchParams,
+    redirectPath: `${PathnameEnum.BUY}/${slug}`,
+  });
+  // The App is rendered by the bridge layout; the page only initializes it.
```
Comment (chrstph-dvx) [inline packages/app/src/app/(with-sidebar)/bridge/buy/[slug]/page.tsx:25]: That seems wrong, page should be rendered by the page itself, not the layout?

packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanel.tsx
```diff
@@ -195,13 +197,13 @@ export function TransferPanel() {
...
 
-  const resetAmountAndSwitchToTransactionHistoryTab = useCallback(() => {
-    setQueryParams({
-      tab: tabToIndex[TabParamEnum.TX_HISTORY],
-      amount: '',
-      amount2: '',
-    });
-  }, [setQueryParams]);
+  const resetAmountAndNavigateToTransactionHistory = useCallback(() => {
+    const newParams = new URLSearchParams(searchParams.toString());
```
chrstph-dvx: NIT: we don't need to cast searchParams to string here, you can pass a URLSearchParams as a parameter of URLSearchParams constructor

packages/app/src/app/(with-sidebar)/bridge/buy/[slug]/page.tsx
Comment (fionnachan) [reply packages/app/src/app/(with-sidebar)/bridge/buy/[slug]/page.tsx:25]: NextJS Pages remount on navigation and so if we want `BridgeClient` to persist across Pages, we need to move it to Layout. But at the same time we need a `page.tsx` to render the route, so we need this page. Layout cannot get `searchParams` so we can't do the `sanitizeAndRedirect` in `initializeBridgePage` on `layout.tsx`

We could consider moving this part to middleware and purely return `null` on this page but then we'll also need to call `registerLocalNetwork` there and it doesn't feel like the right place, plus it's gonna increase the load time

Comment (chrstph-dvx) [reply packages/app/src/app/(with-sidebar)/bridge/buy/[slug]/page.tsx:25]: sanitizeAndRedirect should only be called for bridge, no other route, so it's fine having it only in the bridge route.

> We could consider moving this part to middleware and purely return null on this page 

Ideally, we should not return null in pages, it usually a sign that we're doing something wrong. If all pages render the same content (from layout), then why do we have pages?

Layout should only render the common part (tabs, menu, navigation), and pages should return their content (through children props). /bridge should render `Transfer panel`, /tx-history should render tx history, etc.

I don't think having the page unmounting is an issue, it's actually better for performance. But if we want to keep some of the state, we can use `Activity` (https://react.dev/reference/react/Activity)

In MainContent.tsx:
```

// Determine which content to show based on route and query params
const isBuyPage = showBuyPanel && isBridgeBuyOrSubpages(pathname);
const isTxHistory = searchParams.get('tab') === TabParamEnum.TX_HISTORY;
```

This is the issue, MainContent should not decide what to render. This should be driven from the page itself. This was the idea behind the ticket, defer to the framework to handle this, and not have custom code

PR #85, opened 2025-10-10
feat: deprecate teleport transfers

dewanshparashar:
Completely gets rid of all instances of teleport/layer-leap in codebase 

Test:  check for my address `0x2cd28Cda6825C4967372478E87D004637B73F996` it has a latest teleport txn from Sepolia to Rari (https://sepolia.etherscan.io/tx/0xe18a4567c1d335483c04f085f604cd2c01789b223c2000d6f8d5470d706e7f07) which should not be visible in the tx history at all now.

Closes FS-2237


fionnachan (approved):
good riddance

PR #91, opened 2025-10-13
ci: enable e2es again

fionnachan:


packages/arb-token-bridge-ui/tests/e2e/specs/txHistory.cy.ts
```diff
@@ -31,7 +31,6 @@ describe('Transaction History', () => {
     context('open transactions history panel', () => {
       cy.switchToTransactionHistoryTab('settled');
       cy.findByLabelText('Load More Transactions').click();
-      cy.findAllByTestId(CLAIMABLE_ROW_IDENTIFIER).its('length').should('be.gt', 0);
```
fionnachan: why is this removed?

dewanshparashar: The latest `settled` txns in the test wallet no longer have withdrawal (claimable) txns. Pending tab has a claimable txn, and settled tabs are all deposits. That's why the tests were failing previously.

PR #147, opened 2026-02-03
feat: enhance OFT token detection to block deposits

dewanshparashar:
Redo of https://github.com/OffchainLabs/arbitrum-token-bridge/pull/2537 in the Portal repo
- Updated the API to v2
- Added hardcoded fallback for pyUSD

### DIFF
--- a/packages/arb-token-bridge-ui/src/util/__tests__/WithdrawOnlyUtils.test.ts
+++ b/packages/arb-token-bridge-ui/src/util/__tests__/WithdrawOnlyUtils.test.ts
@@ -1,13 +1,23 @@
 /**
  * @vitest-environment node
  */
-import { describe, expect, it } from 'vitest';
+import { registerCustomArbitrumNetwork } from '@arbitrum/sdk';
+import { beforeAll, describe, expect, it } from 'vitest';
 
 import { ChainId } from '../../types/ChainId';
 import { CommonAddress } from '../CommonAddressUtils';
 import { isWithdrawOnlyToken } from '../WithdrawOnlyUtils';
+import { orbitMainnets } from '../orbitChainsList';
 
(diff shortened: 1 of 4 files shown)


chrstph-dvx (dismissed):

packages/arb-token-bridge-ui/src/util/networks.ts
```diff
@@ -388,7 +388,8 @@ function getIsArbitrumChain(chainId: ChainId) {
   try {
     return !!getArbitrumNetwork(chainId).parentChainId;
   } catch (error) {
-    return false;
+    // Fallback: check if chain is in orbitChains list
+    return !!orbitChains[chainId];
```
dewanshparashar: Updated this to check orbitChains as a fallback when the SDK doesn't recognize the chain. This ensures Orbit chains like Xai are correctly identified and tests pass.

packages/arb-token-bridge-ui/src/util/__tests__/WithdrawOnlyUtils.test.ts
```diff
@@ -0,0 +1,127 @@
...
+      parentChainId: ChainId.ArbitrumOne,
+      childChainId: orbitChainId,
+    });
+    expect(result).toBe(true);
+  });
+
+  it('should allow deposits for USDC.e on a non-Orbit chain', async () => {
+    const result = await isWithdrawOnlyToken({
+      parentChainErc20Address: CommonAddress.ArbitrumOne['USDC.e'],
+      parentChainId: ChainId.Ethereum,
```
fionnachan: this doesn't make any sense because the parent chain is Ethereum so the ERC20 address should be the one on Ethereum instead of Arbitrum One

PR #149, opened 2026-02-05
feat: update nova explorer url

fionnachan:


dewanshparashar (approved):
Approved, but it'd be awesome if this new Explorer link came from constants file instead of being hardcoded everywhere.

packages/portal/common/constants.ts
```diff
@@ -15,6 +15,8 @@ export const SIDEBAR_OPEN_PREFERENCE_KEY = 'arbitrum:portal:sidebar:opened';
 
 export const PORTAL_DATA_ENDPOINT = 'https://portal-data.arbitrum.io';
 
+export const NOVA_EXPLORER_URL = 'https://arbitrum-nova.blockscout.com/';
```
dewanshparashar: Let's remove the `/` at the end otherwise txns might start having double slashes (//) in links

packages/scripts/src/addOrbitChain/schemas.ts
```diff
@@ -39,7 +39,7 @@ export const getParentChainInfo = (parentChainId: number) => {
     case 42170: // Arbitrum Nova
       return {
         rpcUrl: 'https://nova.arbitrum.io/rpc',
-        blockExplorer: 'https://nova.arbiscan.io',
+        blockExplorer: 'https://arbitrum-nova.blockscout.com',
```
dewanshparashar: Not possible to import constant here?

PR #166, opened 2026-02-18
feat: Panax fast withdrawal 15m

fionnachan:


packages/arb-token-bridge-ui/src/util/orbitChainsData.json
```diff
@@ -1420,7 +1420,8 @@
           "name": "Panax",
           "symbol": "PAX",
           "logoUrl": "/images/PanaxMainnet_NativeTokenLogo.svg"
-        }
+        },
+        "fastWithdrawalTime": 900000
```
chrstph-dvx: NIT: I'd change it to 15 * 1000 * 60 (easier to update, and to reason about)

fionnachan: we can't store that unless we want to store it as string

PR #179, opened 2026-02-24
feat: consolidate portal and bridge tooltips

dewanshparashar:
We had 2 different tooltips in the Portal app, both with different styling and using unmaintained Tippy library. This PR fixes that and replaces it with Radix.


packages/app/src/components/Tooltip.tsx
```diff
@@ -15,6 +14,7 @@ export type TooltipProps = {
   onOpenChange?: (open: boolean) => void;
   onClickOutside?: () => void;
 };
+export type TooltipProps = TooltipBaseProps & PropsWithChildren;
```
chrstph-dvx: FYI, you can use PropsWithChildren with generic: PropsWithChildren<TooltipBaseProps>

chrstph-dvx (approved):

PR #238, opened 2026-03-25
feat(bridge): refine assertion interval estimates

douglance:
## Summary

- Add per-chain `assertionIntervalSeconds` to orbit chain configs, computed from on-chain assertion and batch events, replacing the fixed 1-hour buffer in withdrawal time estimates
- Add `update-assertion-intervals` CLI script in `packages/scripts` that queries recent `NodeCreated`/`AssertionCreated` and `SequencerBatchDelivered` events, computes median assertion intervals, and writes results to `orbitChainsData.json`
- Use `assertionIntervalSeconds` in `getConfirmationTime` to give users more accurate withdrawal confirmation estimates

## Details

**Withdrawal time calculation**: `blockTime × confirmPeriodBlocks + assertionIntervalSeconds`

The assertion interval captures the empirical delay between a transaction being sequenced and the assertion covering it being posted on the parent chain. This varies significantly across chains (e.g. Edge Mainnet ~105s vs Molten ~86400s).

**Script behavior**:
- Loads RPC keys from `packages/app/.env` (Alchemy/Infura)
- Processes chains grouped by parent chain with concurrency limit of 3
- Handles per-chain RPC failures gracefully — logs errors, continues processing, writes successful results, exits non-zero if any chain failed
- Chains with insufficient data (<2 assertion events) preserve their existing values
- Computes median of last 20 complete assertion cycles per chain

## Test plan

- [ ] `yarn lint` passes (0 errors)
- [ ] `yarn test:ci` — `WithdrawalUtils.test.ts` and `index.test.ts` pass
- [ ] Run `yarn workspace scripts update-assertion-intervals` and verify output values are reasonable
- [ ] Verify withdrawal confirmation time displays correctly in the bridge UI for orbit chains


packages/scripts/src/updateAssertionIntervals/index.ts
```diff
@@ -0,0 +1,188 @@
...
+
+const ROLLUP_ABI = [
+  'event NodeCreated(uint64 indexed nodeNum, bytes32 indexed parentNodeHash, bytes32 indexed nodeHash, bytes32 executionHash, tuple(tuple(tuple(bytes32[2] bytes32Vals, uint64[2] u64Vals) globalState, uint8 machineStatus) beforeState, tuple(tuple(bytes32[2] bytes32Vals, uint64[2] u64Vals) globalState, uint8 machineStatus) afterState, uint64 numBlocks) assertion, bytes32 afterInboxBatchAcc, bytes32 wasmModuleRoot, uint256 inboxMaxCount)',
+];
+
+const DEFAULT_PARENT_RPCS: Record<number, string> = {
+  1: 'https://eth.llamarpc.com',
+  42161: 'https://arb1.arbitrum.io/rpc',
+  8453: 'https://mainnet.base.org',
+};
```
dewanshparashar: This should come from our existing configured RPCs (in env vars or RPCs defined in code previously as single source of truth)

packages/scripts/src/updateAssertionIntervals/index.ts
```diff
@@ -0,0 +1,188 @@
...
+  const resolvedPath = candidates.find((candidate) => fs.existsSync(candidate));
+
+  if (!resolvedPath) {
+    throw new Error(
+      'Could not locate orbitChainsData.json. Pass the path explicitly with update-assertion-intervals <targetJsonPath>.',
+    );
+  }
+
+  return resolvedPath;
+}
```
dewanshparashar: Why is this needed, ideally we should know exactly where our target file is hosted.

packages/scripts/src/updateAssertionIntervals/index.ts
```diff
@@ -0,0 +1,188 @@
...
+  const rpcUrl = process.env[envKey] ?? DEFAULT_PARENT_RPCS[parentChainId];
+
+  if (!rpcUrl) {
+    throw new Error(`No RPC URL for parent chain ${parentChainId}. Set ${envKey}.`);
+  }
+
+  return new ethers.providers.JsonRpcProvider(rpcUrl);
+}
+
+function roundToNearest30Min(seconds: number): number {
```
dewanshparashar: Please add a comment here - this is not just rounding to 30 mins, but also ensuring that the minimum period is 30mins. If 0 is passed here, the output will be 30mins, not 0.

packages/scripts/src/updateAssertionIntervals/index.ts
```diff
@@ -0,0 +1,188 @@
...
+  return sorted.length % 2 === 0
+    ? (sorted[mid - 1]! + sorted[mid]!) / 2
+    : sorted[mid]!;
+}
+
+function getLookbackBlocks(parentChainId: number): number {
+  const blocksPerDay =
+    parentChainId === 1 ? 7200 : parentChainId === 8453 ? 43200 : 345600;
+
+  return blocksPerDay * 7;
```
dewanshparashar: Should be done using an index, with commented chain names for better clarity.

packages/scripts/src/updateAssertionIntervals/index.ts
```diff
@@ -0,0 +1,188 @@
...
+        chain.assertionIntervalSeconds = interval;
+        updatedChains++;
+        continue;
+      }
+
+      delete chain.assertionIntervalSeconds;
+    }
+  }
+
+  fs.writeFileSync(jsonPath, `${JSON.stringify(data, null, 2)}\n`);
```
dewanshparashar, line 662: This only writes to the file locally. We also need to raise a PR for this change, right?

packages/scripts/src/file-type.d.ts
```diff
@@ -0,0 +1 @@
+declare module 'file-type';
```
dewanshparashar, line 1: Why this change?

dewanshparashar:
+ Harry's comment about making sure that the chains are active, else the calculated assertion times will be quite off from reality.

fionnachan (dismissed):

dewanshparashar (dismissed):

packages/arb-token-bridge-ui/src/util/WithdrawalUtils.ts
```diff
@@ -152,9 +153,11 @@ export function getConfirmationTime(chainId: number) {
     if (blockNumberReferenceChainId === ChainId.Local) {
       confirmationTimeInSeconds = 0;
     } else {
+      const assertionInterval =
+        orbitChains[chainId]?.assertionIntervalSeconds ?? DEFAULT_ASSERTION_INTERVAL_SECONDS;
```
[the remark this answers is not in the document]
douglance: Each bridge operation is a single hop — an L3→Arb One withdrawal settles on Arb One after the L3's challenge period. It doesn't need to wait for Arb One's own assertion to L1. The Arbitrum docs confirm child-to-parent messaging resolves on the parent chain independently of the parent's settlement to its grandparent.

Removed the parent-chain recursion in e86c7bae7 — `getAdditionalWithdrawalDelaySeconds` now just returns the chain's own delay without stacking the parent's.

packages/scripts/src/updateAssertionIntervals/index.ts
douglance: It already does — `loadConfiguredParentRpcUrl` reads the same `NEXT_PUBLIC_ALCHEMY_KEY_*` and `NEXT_PUBLIC_INFURA_KEY_*` env vars the app uses, and only falls back to `getParentChainInfo` as a last resort. Added a JSDoc comment in e86c7bae7 to make this clearer. Also added `dotenv.config()` to load from `packages/app/.env` so the script picks up the keys when run standalone.

packages/scripts/src/updateAssertionIntervals/index.ts
douglance: The target path is a CLI argument — standard for any script that writes to a file. Hardcoding would be fragile (breaks if repo structure changes) and prevents testing with fixture files. The path is always the same in practice (`packages/arb-token-bridge-ui/src/util/orbitChainsData.json`), but keeping it as an arg is the correct design.

packages/scripts/src/updateAssertionIntervals/index.ts
douglance: This function was removed in later commits — the 30-minute rounding was dropped when the estimation approach changed. The script now computes the median of observed assertion intervals directly, without rounding. No action needed.

packages/scripts/src/updateAssertionIntervals/index.ts
douglance: Added inline chain name comments on each case in e86c7bae7: `case 1: // Ethereum Mainnet`, `case 42161: // Arbitrum One`, etc.

packages/scripts/src/updateAssertionIntervals/index.ts
douglance: Correct — the script writes locally. Wrapping it in a GitHub Action that commits and opens a PR is a separate CI concern. Happy to set that up as a follow-up if needed.

packages/scripts/src/file-type.d.ts
douglance: The `file-type` package doesn't ship its own types and `@types/file-type` doesn't exist. Without this ambient module declaration, TypeScript compilation fails. Standard pattern for untyped packages.

douglance:
Re: inactive chains concern — the script already handles this. Chains with fewer than 2 assertion events in the lookback window return `insufficient_data` and their existing values are preserved. We saw this in practice: Unite Mainnet had 0 batch events and was correctly skipped.

douglance:
Resolved all Copilot review comments — they were generated against an earlier revision. The concerns are either already addressed in the current code (block deduplication, `parseAsync`, paged queries, null-safety checks) or not applicable given the input sizes involved (e.g. O(n^3) on ~80 elements).

dewanshparashar (dismissed):

PR #257, opened 2026-04-09
chore: add more code style rules to CLAUDE.md

dewanshparashar:
Adding these rules after learning from another PR review as agents tend to deviate from existing patterns
1. Using `twMerge()`
2. Using `useMemo` cautiously
3. Using typed objects instead of long chain function arguments


chrstph-dvx (approved):
Good call ❤️

douglance (approved):

PR #292, opened 2026-04-21
refactor: new hook for tx panel addPendingTransaction

fionnachan:
Closes FS-2060

### DIFF
--- a/packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
+++ b/packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
@@ -312,6 +312,38 @@ export async function fetchWithdrawalsInBatches(
   return results.flat();
 }
 
+// SWR cache for pending transactions initiated in the current session.
+// Does not fetch; only reads/writes the `new_tx_list` cache. Use this instead
+// of `useTransactionHistory` when a component only needs to add a pending tx.
+export const useNewTransactions = (address: Address | undefined) => {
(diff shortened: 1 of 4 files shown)


packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -312,6 +312,38 @@ export async function fetchWithdrawalsInBatches(
   return results.flat();
 }
 
+// SWR cache for pending transactions initiated in the current session.
+// Does not fetch; only reads/writes the `new_tx_list` cache. Use this instead
+// of `useTransactionHistory` when a component only needs to add a pending tx.
+export const useNewTransactions = (address: Address | undefined) => {
```
dewanshparashar: Why call it `useNewTransactions()` instead of `usePendingTransactions()` or `useAddPendingTransactions()`

fionnachan: pending transactions makes it sound like all pending transactions, and that would include transactions not newly submitted right now on the UI

but if it makes more sense to keep the old name i'll update
wdyt?

dewanshparashar: It's meant as a nit since even this hook may be temporary in the longer run. I would still name it as `useAddPendingTransactions` to be absolutely spot on.

dewanshparashar (approved):

PR #306, opened 2026-05-06
fix(bridge): enfore initialization of orbit chains

### LINKED ISSUE #303: [bug]: destinationChain URL param ignored in bridge page
### Contact Details

dev_hun@aergo.io

## Description

The `destinationChain` query parameter on the Arbitrum Bridge URL is currently being ignored.

Previously, opening a URL like this:

```txt

https://portal.arbitrum.io/bridge?destinationChain=hpp-mainnet&destinationToken=0xe33fbe7584eb79e2673abe576b7ac8c0de62565c

```

correctly preselected the destination chain as `hpp-mainnet`.

However, currently the bridge defaults to `Arbitrum One` instead of the specified destination chain.

This issue does not only affect HPP Mainnet — other custom Orbit chains also appear to have the same problem.

---

## Expected Behavior

The bridge should respect the `destinationChain` query parameter and automatically select the specified chain.

Example:

- `destinationChain=hpp-mainnet`

→ Destination chain should be set to `hpp-mainnet`

---

## Actual Behavior

The destination chain is ignored and defaults to `Arbitrum One`.

---

## Steps to Reproduce

1. Open:

```txt

https://portal.arbitrum.io/bridge?destinationChain=hpp-mainnet&destinationToken=0xe33fbe7584eb79e2673abe576b7ac8c0de62565c

```

2. Check the "To" network field

3. Observe that it shows `Arbitrum One` instead of `hpp-mainnet`

---

## Notes

This seems related to URL parameter parsing or chain preset initialization regression.

The issue appears reproducible across multiple Orbit/custom chains, not only HPP.

---

### What devices are you seeing the problem on?

Desktop

### What browsers are you seeing the problem on?

Chrome

### Browser Version

147.0.7727.56

### Relevant log output

```shell

```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

chrstph-dvx:

### DIFF
--- a/packages/arb-token-bridge-ui/src/util/networks.ts
+++ b/packages/arb-token-bridge-ui/src/util/networks.ts
@@ -490,6 +490,25 @@ export function mapCustomChainToNetworkData(chain: ChainWithRpcUrl) {
   explorerUrls[chain.chainId] = chain.explorerUrl;
 }
 
+let bridgeNetworksInitialized = false;
+
+export function initializeBridgeNetworks() {
+  if (bridgeNetworksInitialized) {
+    return;
+  }
+
+  [...Object.values(orbitChains), ...getCustomChainsFromLocalStorage()].forEach((chain) => {
+    try {
+      registerCustomArbitrumNetwork(chain);
+      mapCustomChainToNetworkData(chain);
+    } catch (_) {
+      // already added
+    }
+  });
+
+  bridgeNetworksInitialized = true;
+}
+
 export function isArbitrumChain(
   chain: BlockNumberReferenceNetwork | ArbitrumNetwork,
 ): chain is ArbitrumNetwork {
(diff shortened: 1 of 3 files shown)


packages/arb-token-bridge-ui/src/util/networks.ts
```diff
@@ -490,6 +490,25 @@ export function mapCustomChainToNetworkData(chain: ChainWithRpcUrl) {
   explorerUrls[chain.chainId] = chain.explorerUrl;
 }
```
dewanshparashar, line 492: Why did we place this in networks.ts and not in initialization.ts where a similar `addOrbitChainsToArbitrumSDK` is also defined

dewanshparashar (approved):

PR #311, opened 2026-05-08
fix(bridge): handle token withdrawals without `TransferRouted` event

dewanshparashar:
We use `TransferRouted` event to extract `sender` and `destination` for a withdrawal - but this lengthy workaround was only valid for SCW transactions. For EOA's the `from` and `to` data is already available in the withdrwal-initiated event log. 
So for normal transctions, we use the readily available info, and only when we find missing data, we try extracting it through `TransferRouted` logs.

Test for `0x3844597c2b75E8510d8895173adDC28dd995fA45`

### DIFF
--- a/packages/arb-token-bridge-ui/src/util/withdrawals/helpers.ts
+++ b/packages/arb-token-bridge-ui/src/util/withdrawals/helpers.ts
@@ -199,35 +199,43 @@ export async function mapTokenWithdrawalFromEventLogsToL2ToL1EventResult({
     result.childChainId,
   );
 
-  // We cannot access sender and destination from the withdrawal object.
-  // We have to get them from the receipt logs.
-  //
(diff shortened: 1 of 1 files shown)


packages/arb-token-bridge-ui/src/util/withdrawals/helpers.ts
```diff
@@ -199,35 +199,43 @@ export async function mapTokenWithdrawalFromEventLogsToL2ToL1EventResult({
     result.childChainId,
   );
 
-  // We cannot access sender and destination from the withdrawal object.
```
dewanshparashar, line 202: Just moving this whole block inside `if (!sender || !destinationAddress)` was the core fix.

fionnachan (approved):

PR #318, opened 2026-05-13
refactor(earn): use Zerion for asset prices

dewanshparashar:
Closes FS-1992


packages/app/src/app/api/onchain-actions/v1/earn/lib/zerionService.test.ts
```diff
@@ -0,0 +1,256 @@
...
+
+  afterEach(() => {
+    if (originalApiKey === undefined) {
+      delete process.env.ZERION_API_KEY;
+    } else {
+      process.env.ZERION_API_KEY = originalApiKey;
+    }
+    global.fetch = originalFetch;
+    vi.restoreAllMocks();
+  });
```
chrstph-dvx, line 45: That seems brittle. Can we pass fetch as a dependency maybe to `fetchZerionCurrentPrices`? (and make it optional, by setting the default value to global.fetch)?

packages/app/src/app/api/onchain-actions/v1/earn/lib/zerionService.ts
```diff
@@ -0,0 +1,284 @@
...
+          }
+        }
+      }),
+    );
+  }
+
+  const settled = await Promise.allSettled(requests);
+  for (const outcome of settled) {
+    if (outcome.status === 'rejected') {
+      console.warn('[earn][zerion] batched price fetch failed:', getErrorMessage(outcome.reason));
```
chrstph-dvx, line 167: If it used on the frontend part, we should use the logger instead

dewanshparashar: This is used server-side. Logger is window aware, won't work here.

chrstph-dvx (approved):

PR #322, opened 2026-05-14
fix(bridge): optimize mind network tx history fetch

dewanshparashar:
## Summary

- Add Mind Network (228) to `BATCH_FETCH_BLOCKS` at 10k — Mind's RPC rejects larger `eth_getLogs` windows.
- The 10k batch size meant `fetchWithdrawals` ran ~425× per page load. For addresses with zero nonce on the chain, every batch wastefully fired `eth_getTransactionCount` + `/block-number` + subgraph calls. Added an early-exit in `fetchWithdrawalsInBatches`: when the sender has zero nonce on an Orbit chain and `forceFetchReceived` is false, return `[]` before slicing. `forceFetchReceived` continues to opt out (SCW flows).

Check for `0x1BD4a8FcaFa8e16c6DF3C5C9E2359beF2D03A807`.

Fixes SUP-1391.

## Test plan
- [ ] Open tx history on Mind Network with the test address — withdrawals load without RPC errors.
- [ ] Address with nonce > 0 on an Orbit chain still triggers full fetch.
- [ ] SCW flow (forceFetchReceived = true) still triggers full fetch.


chrstph-dvx (dismissed):

fionnachan (approved):
LGTM, thank you!

PR #323, opened 2026-05-15
perf(bridge): skip pre-history in batched withdrawal fetch

dewanshparashar:
## Why

#322 stopped Mind Network from blowing up the RPC with oversized `eth_getLogs` queries — but the fix slices the chain into 10k-block batches, and Mind is now at ~59.7M blocks. For users with any L3 activity on Mind, every page-load still fires ~6,000 `fetchWithdrawals` invocations → **~18,000 network calls**  and **~20 minutes** to fully resolve. The vast majority of those batched-withdrawal fetches return zero - because the user can't have withdrawn before their first L3 tx ever landed.

## Change

Binary-search for the lowest block where the sender has nonce > 0 (their first L3 tx) and use it as the `fromBlock` for batching. ~26 sequential `eth_getTransactionCount` calls upfront, but cuts thousands of pointless batches.

Gated to:
- Orbit chains where `senderNonce > 0` (the existing `senderNonce === 0` early-return handles zero-nonce users)
- Self-withdrawals (`sender == receiver`) — same heuristic as the existing receiver gate
- Small batch sizes (`≤ 100k`) — skips ApeChain (5M batch, ~8 total batches) where the probe cost outweighs the savings

Falls back to `fromBlock = 0` if the RPC rejects historical nonce queries.

## Performance impact

**Mind Network** (chain 228, ~59.7M blocks, 10k batch size):

| Scenario | Before any fix | After #322 | After this PR |
|---|---|---|---|
| Zero-nonce user (no L3 activity) | ~3 calls / instant | **1 call / instant** | 1 call / instant |
| Active user (has L3 activity) | Tx history breaks - RPC returns query too large | ~18,000 calls / ~20 min | **~3,000 calls¹ / ~60s** |

¹ Scales with how recently the user first transacted on Mind. Most users will land between 1,500–5,000 calls.

**Other batched Orbit chains**:

| Chain | Before #322 | After #322 | After this PR |
|---|---|---|---|
| T-REX (1628, 1.28M blocks, 10k batch) | ~384 calls | unchanged | **~100–380 calls** |
| ApeChain (33139, 38M blocks, 5M batch) | ~24 calls | unchanged | unchanged (intentionally excluded — probe cost would outweigh savings) |

## Test plan
- [ ] Test for `0x1BD4a8FcaFa8e16c6DF3C5C9E2359beF2D03A807` (wallet with mind network activity) and see the difference between this vs prod
- [ ] Tx history experience for the rest of the use cases remains unchanged

### DIFF
--- a/packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
+++ b/packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
@@ -315,7 +315,10 @@ export async function fetchWithdrawalsInBatches(
       typeof params.batchSizeBlocks === 'number' &&
       params.batchSizeBlocks <= SMALL_BATCH_SIZE_THRESHOLD;
 
-    if (batchSizeIsSmall && params.receiver && addressesEqual(params.sender, params.receiver)) {
+    // Apply when there are no separate receiver queries to worry about:
+    // either receiver is undefined (e.g., SCW connected to child chain — only
+    // sent txns are fetched), or receiver == sender (EOA self-withdrawal).
+    if (batchSizeIsSmall && (!params.receiver || addressesEqual(params.sender, params.receiver))) {
       const firstBlock = await findFirstBlockWithNonce({
         address: params.sender,
         provider: params.l2Provider,
--- a/packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
+++ b/packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
@@ -322,7 +322,7 @@ export async function fetchWithdrawalsInBatches(
         latestBlock: latestBlockNumber,
       });
       if (firstBlock > fromBlock) {
-        fromBlock = firstBlock;
+        fromBlock = Math.max(0, firstBlock - 1);
       }
     }
   }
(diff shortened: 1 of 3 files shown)


packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -300,6 +300,31 @@ export async function fetchWithdrawalsInBatches(
...
+    // Only do this when BATCH_FETCH_BLOCKS is small. On chains like ApeChain
+    // (5M batch size, ~8 total batches), the ~25 sequential probe calls cost
+    // more than they save. On Mind/T-REX (10k batch size, hundreds/thousands
+    // of batches), the savings dominate.
+    const SMALL_BATCH_SIZE_THRESHOLD = 100_000;
+    const batchSizeIsSmall =
+      typeof params.batchSizeBlocks === 'number' &&
+      params.batchSizeBlocks <= SMALL_BATCH_SIZE_THRESHOLD;
+
+    if (batchSizeIsSmall && params.receiver && addressesEqual(params.sender, params.receiver)) {
```
fionnachan: the `addressesEqual` comparison is not very accurate because `params.sender` or `params.receiver` will be `undefined` when we don't want to include txs when the address is the sender or recipient respectively

so even if `params.sender` is `undefined`, it could still mean that the recipient and the sender are the same address

Image

fionnachan (approved):
LGTM!

PR #324, opened 2026-05-15
feat(earn): usd pricing for assets

dewanshparashar:
## Summary

Fixes USD pricing in Earn end-to-end. Structured in three layers for review:

**1. API + adapters** — `StandardOpportunityBase` carries `underlyingTokenPriceUsd` / `shareTokenPriceUsd`; `StandardUserPosition` carries `tokenPriceUsd`. LiFi (Zerion), Pendle (Pendle prices), and Vaults adapters populate them. Price-endpoint failures degrade to `null` and never block opportunity rendering.

**2. Hook (for easy integration)** — new `useEarnPrices` purely derives `priceByAddress` + `priceBySymbol` maps from already-fetched opportunities and positions (no extra fetch). `useEarnTokenPrice` wrapper falls back to `useETHPrice` for native ETH.

**3. UI integration** — same hook consumed across all three opportunity panels (`VaultActionPanel`, `LiquidStakingActionPanel`, `PendleActionPanel`), plus tx history + details. Shared formatter `getEarnInputUsdValue` keeps the input → USD conversion consistent.

<p float="left">
image
image
image
</p>

## Test plan
- [ ] Lend: input shows live USD valuation.
- [ ] Liquid staking: USD shown on both input + receive sides; native ETH falls back to `useETHPrice`.
- [ ] Pendle: PT and underlying both priced.
- [ ] Tx history: USD resolves for all three categories, including Pendle `PT*` rows.
- [ ] Force a price-endpoint failure → rows still render, `valueUsd: null`.

Closes FS-2094

### DIFF
--- a/packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
+++ b/packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
@@ -604,3 +604,51 @@ export async function getPendleTransactionHistory({
     pageSize: rawData.pageSize || pageSize,
   };
 }
+
+/**
+ * Batched USD price lookup for Pendle assets (PT, SY, YT, underlying).
+ * Returns a Map keyed by lowercased address. Missing entries map to null.
+ * Docs: https://api-v2.pendle.finance/core/docs#tag/assets/get/v1/prices/assets
+ */
+export async function getPendleAssetPrices(params: {
+  chainId: number;
+  addresses: string[];
+}): Promise<Map<string, number | null>> {
+  const { chainId, addresses } = params;
+  const result = new Map<string, number | null>();
+  if (addresses.length === 0) return result;
+
+  const unique = Array.from(new Set(addresses.map((a) => a.toLowerCase())));
+  for (const addr of unique) result.set(addr, null);
+
+  const url = new URL(`${PENDLE_API_BASE_URL}/v1/${chainId}/assets/prices`);
+  url.searchParams.set('addresses', unique.join(','));
+
+  const response = await fetch(url.toString());
+  if (!response.ok) throw new Error(await parsePendleApiError(response));
+
+  const payload = (await response.json()) as
+    | { pricesUsd?: Record<string, number | string> }
+    | { prices?: Record<string, number | string> }
+    | { data?: Array<{ address?: string; priceUsd?: number | string; price?: number | string }> };
+
+  const priceMap =
+    ('pricesUsd' in payload && payload.pricesUsd) ||
+    ('prices' in payload && payload.prices) ||
+    null;
+
+  if (priceMap) {
+    for (const [addr, raw] of Object.entries(priceMap)) {
+      const num = Number(raw);
+      result.set(addr.toLowerCase(), Number.isFinite(num) ? num : null);
+    }
+  } else if ('data' in payload && Array.isArray(payload.data)) {
+    for (const item of payload.data) {
+      if (!item?.address) continue;
+      const num = Number(item.priceUsd ?? item.price);
+      result.set(item.address.toLowerCase(), Number.isFinite(num) ? num : null);
+    }
+  }
+
+  return result;
+}
--- a/packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
+++ b/packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
@@ -605,7 +605,8 @@ export async function getPendleTransactionHistory({
   };
 }
 
-// GET /v1/{chainId}/assets/prices?addresses=0x... → { prices: { "0x...": number } }
+// Docs: https://api-v2.pendle.finance/core/docs#tag/assets/get/v1/prices/assets
+// Response keys are "{chainId}-{address}"; we strip the prefix so consumers look up by plain address.
 export async function getPendleAssetPrices(params: {
   chainId: number;
   addresses: string[];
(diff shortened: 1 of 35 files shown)


packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
```diff
@@ -604,3 +604,33 @@ export async function getPendleTransactionHistory({
     pageSize: rawData.pageSize || pageSize,
   };
 }
+
+// Docs: https://api-v2.pendle.finance/core/docs#tag/assets/get/v1/prices/assets
```
fionnachan, line 622: chainId should be a query param according to this doc, so we should implement that instead?

packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
```diff
@@ -604,3 +604,33 @@ export async function getPendleTransactionHistory({
...
+  for (const addr of unique) result.set(addr, null);
+
+  const url = new URL(`${PENDLE_API_BASE_URL}/v1/${chainId}/assets/prices`);
+  url.searchParams.set('addresses', unique.join(','));
+
+  const response = await fetch(url.toString());
+  if (!response.ok) throw new Error(await parsePendleApiError(response));
+
+  const payload = (await response.json()) as {
+    pricesUsd?: Record<string, number | string>;
```
fionnachan: where does this shape come from?
because I don't see `pricesUsd` from the response
https://api-v2.pendle.finance/core/v1/42161/assets/prices?skip=0&type=PENDLE_LP

https://api-v2.pendle.finance/core/v1/prices/assets?skip=0&type=PENDLE_LP&chainId=1

packages/app/src/hooks/earn/useEarnPrices.ts
```diff
@@ -0,0 +1,124 @@
...
+      const underlyingPrice = parseFiniteNumber(row.underlyingTokenPriceUsd, { min: 0 });
+      const sharePrice = parseFiniteNumber(row.shareTokenPriceUsd, { min: 0 });
+
+      if (row.underlyingTokenAddress && underlyingPrice) {
+        byAddress.set(makeAddressKey(row.chainId, row.underlyingTokenAddress), underlyingPrice);
+      }
+      if (row.shareTokenAddress && sharePrice) {
+        byAddress.set(makeAddressKey(row.chainId, row.shareTokenAddress), sharePrice);
+      }
+      if (row.token && underlyingPrice) {
```
fionnachan, line 68: does this cover multiple pendle tokens for the same underlying token with different maturity dates?

packages/app/src/components/earn/EarnTransactionHistoryTable.tsx
```diff
@@ -132,6 +132,26 @@ function getDisplayAsset(
...
+      };
+    }
+    if (hasOutputAsset) {
+      return {
+        amountRaw: row.outputAssetAmountRaw!,
+        symbol: row.outputAssetSymbol!,
+        decimals: row.outputAssetDecimals!,
+        logo: row.outputAssetLogo ?? row.assetLogo,
+      };
+    }
```
chrstph-dvx: Can we either improve the type, so we know that when action === 'sell' then we're guaranteed to have those input properties (and if it's different than sell, then we have output... properties), or check manually in the if/else here?

packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
dewanshparashar: Good catch - this was overwritten when I was trying out something else. Fixed the `prices` key.

packages/app/src/app/api/onchain-actions/v1/earn/lib/pendleApi.ts
dewanshparashar: Updated the implementation to reflect the latest v2 pendle price api

packages/app/src/components/earn/EarnTransactionHistoryTable.tsx
dewanshparashar: refactored the if-else to make it less verbose and better for readability

packages/app/src/hooks/earn/useEarnPrices.ts
dewanshparashar: The get price-by-symbol isn't. That's why we also have a get-price-by-address alternative exposed by the same hook which can be used for colliding cases since address will be unique across pendle markets with different maturity dates.

packages/app/src/hooks/earn/useEarnPrices.ts
```diff
@@ -0,0 +1,124 @@
...
+      // Pendle's tx history uses the "PT"+underlying convention as the row symbol.
+      if (row.category === OpportunityCategory.FixedYield && row.token && sharePrice) {
+        bySymbol.set(makeSymbolKey(row.chainId, `PT${row.token}`), sharePrice);
+      }
+    }
+
+    for (const position of positionsMap.values()) {
+      const positionPrice = parseFiniteNumber(position.tokenPriceUsd, { min: 0 });
+      if (!position.tokenAddress || !positionPrice) continue;
+      const key = makeAddressKey(chainId, position.tokenAddress);
```
fionnachan, line 80: can you double check this for pendle please?

https://api-v2.pendle.finance/core/docs#tag/markets/get/v2/markets/all

the `position.tokenAddress` would be sth like this `1-0xb87511364014c088e30f872efc4a00d7efb843ac` instead of `0xb87511364014c088e30f872efc4a00d7efb843ac` so could be an issue?

dewanshparashar: Although this wasn't impacting the functionality since the same composite id was being used to reference the token in all Pendle flows, but at the core it should be consistent. So I normalized the pendle token address at the source, and now it's being reflected as a plain address everywhere downstream as intended.

chrstph-dvx (approved):

fionnachan (approved):

PR #337, opened 2026-05-25
fix(earn): clean up API failure handling

dewanshparashar:
## Summary

Cleanup of failures in the earn API that we found in server logs. Three changes:

1. **Empty-result quotes return 200, not errors** — LiFi `NO_ROUTES_FOUND` and Pendle `QUOTE_ROUTE_NOT_FOUND` were really "computation produced no executable route," which is a valid success path, not a server error.
2. **Validation errors get proper HTTP status + code** — `userAddress is required` from Vaults now flows through `ValidationError` → 400, not a bare `Error` → 500. Also - we are no longer fetching quotes for Lending markets (vaults) if the wallet is not connected - this takes care of this error from the UI side.
3. **Log level follows status** — route handlers now `console.warn` for 4xx and `console.error` for 5xx, so validation failures stop tripping the 5xx alert rules.

## Behavioral change to flag

`/transaction-quote` consumers previously got 5xx/4xx with `code: NO_ROUTES_FOUND` or `QUOTE_ROUTE_NOT_FOUND` when no route existed. They now get a successful 200 response with `canExecute: false` and `transactionSteps: []`. Internal callers (frontend) already treat empty steps as "not ready," so no UI changes were required.

## Test plan

- [ ] Manually verify Vaults supply / redeem still works end-to-end with a connected wallet
- [ ] Manually verify Pendle and Liquid Staking quotes still work with a connected wallet
- [ ] Trigger the no-route case on Liquid Staking (e.g., bogus token pair) and confirm 200 + empty steps, no error toast, submit disabled
- [ ] Confirm disconnected Vaults panel no longer triggers a 400 in network tab (no quote request fired)
- [ ] Monitor 5xx error rates on the earn endpoints after deploy

Closes FS-2236

### DIFF
--- a/packages/app/src/app/api/onchain-actions/v1/earn/adapters/PendleAdapter.ts
+++ b/packages/app/src/app/api/onchain-actions/v1/earn/adapters/PendleAdapter.ts
@@ -291,7 +291,7 @@ export class PendleAdapter implements VendorAdapter {
     id: string,
     request: TransactionQuoteRequest,
     chainId: EarnChainId = DEFAULT_CHAIN_ID,
-  ): Promise<TransactionQuoteResponse> {
+  ): Promise<TransactionQuoteResponse | null> {
     const {
       action,
       amount,
@@ -366,15 +366,7 @@ export class PendleAdapter implements VendorAdapter {
 
     const firstRoute = route.routes?.[0];
     if (!firstRoute) {
-      return {
-        opportunityId: id,
-        vendor: Vendor.Pendle,
-        action,
-        canExecute: false,
-        estimatedGas: '0',
-        estimatedGasUsd: '0',
-        transactionSteps: [],
-      };
+      return null;
     }
 
     const transactionSteps: TransactionStep[] = [];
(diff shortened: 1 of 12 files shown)


packages/app/src/app/api/onchain-actions/v1/earn/adapters/PendleAdapter.ts
```diff
@@ -366,10 +366,15 @@ export class PendleAdapter implements VendorAdapter {
 
     const firstRoute = route.routes?.[0];
     if (!firstRoute) {
-      throw new ValidationError(
-        'QUOTE_ROUTE_NOT_FOUND',
-        'No route found in Pendle convert response',
-      );
+      return {
```
chrstph-dvx: (same for LifiAdapter), if we found no route, what about returning null rather than a "mocked" route? So it's easier to reason about from the frontend PoV

dewanshparashar: Good idea. Made the change.

chrstph-dvx (approved):

fionnachan (approved):

PR #338, opened 2026-05-25
fix(earn): APY discrepancies in UI

dewanshparashar:
## Summary

Post-launch fixes for the Earn feature based on Slack feedback:

- Lend (Vaults) APY switched from 30-day → 7-day, matching what users see on vaults.fyi (and much more stable than showing the fluctuating 1d APY)
- wstETH switched from Lido's instant APR → 7-day moving average for parity with weETH
- Liquid staking (wstETH, weETH) now converts the Dune 7-day moving-average **APR → APY** via daily compounding (Dune only exposes APR), so the value matches the "APY" label. Shared `convertAprToApy` helper in `NumberUtils` with unit tests.
- "7-day average APY" tooltip on Lend & Liquid Staking column headers
- Y-axis labels in historical charts no longer crop (e.g. `0.051%`)
- `formatPercentage` now trims trailing zeros (e.g. `0.050%` → `0.05%`) so APY rendering is consistent across rows, cards, chart tooltips, and breakdowns
- Realigned the APY tooltip to the new Radix `Tooltip` API (post Next upgrade)

## Test plan

- [x] /earn landing — Lend & LS sections show 7-day average APYs with info tooltip
- [x] /earn opportunity detail (Aave WETH / WBTC) — y-axis labels not cropped on APY chart
- [x] Liquid staking APY reflects APR→APY conversion (wstETH ~2.39% APR → ~2.42% APY)
- [ ] All APY values (rows, cards, chart tooltip, breakdown) use consistent decimals with no trailing zeros

Closes FS-2235

### DIFF
--- a/packages/app/src/app/api/onchain-actions/v1/earn/lib/duneService.ts
+++ b/packages/app/src/app/api/onchain-actions/v1/earn/lib/duneService.ts
@@ -53,6 +53,7 @@ function findColumn(
 
 /** APY columns that are already expressed as percentages (not decimals). */
 const PERCENTAGE_APY_MATCHERS: Array<(name: string) => boolean> = [
+  (n) => n === 'lido staking apr (ma_7)' || n === 'lido staking apr(ma_7)',
   (n) => n === 'lido staking apr (instant)' || n === 'lido staking apr(instant)',
   (n) => n === 'steth apy',
 ];
@@ -96,12 +97,16 @@ function transformDuneResults(rows: Array<Record<string, unknown>>): DuneDataPoi
 
   const apyColumn = findColumn(
     allColumns,
-    // wstETH new query (already percentage)
+    // wstETH new query — 7-day moving average (already percentage)
+    (n) => n === 'lido staking apr (ma_7)' || n === 'lido staking apr(ma_7)',
+    // wstETH new query — instant fallback (already percentage)
     (n) => n === 'lido staking apr (instant)' || n === 'lido staking apr(instant)',
     // wstETH old query (already percentage)
     (n) => n === 'steth apy',
-    // weETH (decimal format)
-    (n) => n === 'avg_7day_apr' || n === 'daily_apr',
+    // weETH 7-day average (decimal format)
+    (n) => n === 'avg_7day_apr',
+    // weETH daily fallback (decimal format)
+    (n) => n === 'daily_apr',
     // Generic fallback
     (n) => n.includes('apy') || n.includes('apr') || n.includes('yield'),
   );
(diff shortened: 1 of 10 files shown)


chrstph-dvx (dismissed):

fionnachan (commented):
The Aave APY are still different. Seems like the one on our UI is APR instead of APY. I looked at Compound and it was different as well. You might want to check if the rate returned for the lending protocols are APR (without the compounding effect) instead of APY.

Image

Image

Image

Image


weETH has a different APY too

Image

Image

packages/app/src/app/api/onchain-actions/v1/earn/lib/duneService.ts
```diff
@@ -53,6 +53,7 @@ function findColumn(
 
 /** APY columns that are already expressed as percentages (not decimals). */
 const PERCENTAGE_APY_MATCHERS: Array<(name: string) => boolean> = [
+  (n) => n === 'lido staking apr (ma_7)' || n === 'lido staking apr(ma_7)',
```
fionnachan, line 57: hmmm also here. shouldn't put APR as APY because they are different, might need to convert it with a small calculation in our codebase if you can't get the APY from Dune

packages/arb-token-bridge-ui/src/util/NumberUtils.ts
```diff
@@ -38,6 +38,7 @@ export function formatCompactNumber(n: number): string {
 }
 
 export function formatPercentage(value: number): string {
+  if (value === 0) return '0%';
   if (value < 0.01) return `${value.toFixed(4)}%`;
   if (value < 1) return `${value.toFixed(3)}%`;
   return `${value.toFixed(2)}%`;
```
fionnachan: `toFixed()` pads 0 to values with 1 fixed point and it is not based on the number of decimals so we got 0.050% which is probably not what we want to show on Earn

Image

packages/arb-token-bridge-ui/src/util/NumberUtils.ts
```diff
@@ -38,9 +38,11 @@ export function formatCompactNumber(n: number): string {
 }
 
 export function formatPercentage(value: number): string {
-  if (value < 0.01) return `${value.toFixed(4)}%`;
-  if (value < 1) return `${value.toFixed(3)}%`;
-  return `${value.toFixed(2)}%`;
+  if (value === 0) return '0%';
+  // `Number` drops the trailing zeros `toFixed` pads (e.g. 0.05 -> "0.050").
```
chrstph-dvx, line 42: What about 5.05? Will it format to 5.050?

dewanshparashar: This is already taken care of by the `Number()` wrap, but added a test for the case in comment.

fionnachan (approved):

chrstph-dvx (approved):

PR #341, opened 2026-05-27
build(deps): bump react-toastify

chrstph-dvx:


fionnachan (approved):
tested, works well thank you!

PR #343, opened 2026-05-27
fix(bridge): dedupe transaction in tx history

chrstph-dvx:
### Summary

Transaction was added pending after transfer, and it was then later picked up from RPC with other transactions. We're now deduping them (tx from RPC has a higher priority)


packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -269,6 +269,49 @@ function dedupeTransactions(txs: Transfer[]) {
...
+  fetchedTransactions?: MergedTransaction[][];
+}) {
+  const flattenedFetchedTransactions = fetchedTransactions.flat();
+  const fetchedByIdentity = new Map<string, MergedTransaction>();
+
+  for (const tx of flattenedFetchedTransactions) {
+    if (!isTransactionForAddress(tx, address)) {
+      continue;
+    }
+    fetchedByIdentity.set(getMergedTransactionIdentity(tx), tx);
```
fionnachan: This change introduces a regression because it would treat batched withdrawals as the same tx and deduplicated them.

Try with this wallet address `0x5eB159b3E6271E6993318c2EE63Fe259Cb0e9f1E`. Support was introduced by #226.

Image

On this branch, it became:
Image

which means from now on, if a user submitted a batched withdrawal using Safe, they won't be able to claim each individual withdrawal using our UI.

Please fix this and add a test case to cover it.

fionnachan (approved):
LGTM thanks!

PR #357, opened 2026-06-16
ci: use arbitrum-testnode service container

douglance:
## Summary

Replaces the slow `OffchainLabs/actions/run-nitro-test-node` composite action (which builds the testnode from a git ref on every job, ~20 min/job) with the **prebuilt `arbitrum-testnode-ci` service-container image**, pinned to the latest version with a complete variant set: **`v0.2.2-nc3.2`**.

## Changes

- **`e2e-tests.yml`**: `test-e2e` now runs inside a `cypress/browsers` container with the testnode as a `services:` container (addressed as `testnode`). Adds steps to wait for testnode health, fetch `localNetwork.json` into `networksNitroTestnode.generated.json`, and port-forward RPC ports for the app.
- **`formatSpecfiles.js`**: adds a `tagMap` so each matrix variant selects the right image tag (`regular`→`l2`, `orbit-eth`→`l3-eth`, custom 6/18/20→`l3-custom-*`).
- **`networksNitroTestnode.ts`**: reads the fetched testnode config instead of constructing it manually (−95/+11).
- Splits CCTP into a dedicated `test-e2e-cctp` job (invoked by `run-cctp-tests.yml`), which doesn't need the testnode service.

## Validation

- Workflow YAML parses; `formatSpecfiles.js` yields 65 regular entries with correct tags + 1 cctp entry
- `networksNitroTestnode.generated.json` is valid JSON and consumed by the app


packages/arb-token-bridge-ui/tests/support/common.ts
```diff
@@ -282,24 +290,60 @@ export async function checkForAssertions({
...
+  }
+
+  if (isBold) {
+    const boldContract = new ethers.Contract(rollupAddress, boldAbi, parentProvider);
+    /* eslint-disable no-await-in-loop */
+    while (true) {
+      try {
+        const latestConfirmed = await boldContract.latestConfirmed();
+        const assertion = await boldContract.getAssertion(latestConfirmed);
+        const createdAtBlock = assertion.createdAtBlock.toNumber();
```
fionnachan, line 323: i think this is causing some of the tests to fail

Image

fionnachan (approved):
LGTM, amazing!

PR #360, opened 2026-06-17
fix(bridge): batch transfers updates

dewanshparashar:
Batch transfers (token + ETH in the same tx) were showing LiFi/Across/Mayan routes in the route list, but those only bridge the token and leave the ETH behind. The canonical Arbitrum route is the only one that can actually move both, since the extra ETH goes in as the retryable's L2 callvalue.

image

Couple of changes here:

- When amount2 > 0, only fetch and show the canonical route. We skip the LiFi/CCTP/OFT quotes entirely so you can't pick a route that drops the ETH leg.
- Auto-expand the second input when amount2 comes in from the URL. That effect was living inside Input2, which only renders once the input is already open, so it never fired for deep links / shared URLs.

### DIFF
--- a/packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanelMain/SourceNetworkBox.tsx
+++ b/packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanelMain/SourceNetworkBox.tsx
@@ -162,8 +162,6 @@ const Input1 = React.memo(() => {
 Input1.displayName = 'Input1';
 
 const Input2 = React.memo(() => {
-  const { showAmount2Input } = useAmount2InputVisibility();
-
   const [networks] = useNetworks();
   const { childChainProvider } = useNetworksRelationship(networks);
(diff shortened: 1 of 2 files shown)


packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanelMain/SourceNetworkBox.tsx
```diff
@@ -183,12 +180,6 @@ const Input2 = React.memo(() => {
     }
   }, [amount2, maxAmount2, isMaxAmount2, setAmount2]);
 
-  useEffect(() => {
```
dewanshparashar, line 186: Moved out of `Input2` which is conditionally rendered (and hence misses opening itself when page mounts) to the parent component which is always mounted and decides whether `Input2` should open on deep-link or not.

fionnachan (approved):
LGTM

PR #364, opened 2026-06-19
feat(bridge): backfill CCTP history from chain logs

dewanshparashar:
## Problem

CCTP transaction history is read from a subgraph. Its `messageSent` entity is served by The Graph's decentralized network, where the gateway routes each query to a different indexer — and those indexers are intermittently out of sync. So a user's recent CCTP transfer is **present on some requests and missing on others**, seconds apart.

Reproduced on prod for an affected wallet: the transfer showed in only **3 of 8** spaced calls (~38%). Users currently work around it by refreshing several times.

## Fix

On each CCTP history request, re-read the recent window (~1 week) directly from chain logs (`MessageSent` on the source `MessageTransmitter`, `MessageReceived` on the target) and merge it with the subgraph results:

- deduped by tx hash (subgraph wins),
- **fail-open** — any backfill error returns the subgraph-only response unchanged,
- enable-gated via `CCTP_GAP_BACKFILL_ENABLED`.

History becomes deterministic regardless of which indexer answered.

## Changes

- **`util/cctp/cctpBackfill.ts`** (new) — `fetchCctpGapBackfill` reads recent CCTP burns/mints and reshapes them to the subgraph types; `mergeCctpBackfill` orchestrates fetch + dedupe + fail-open.
- **`app/api/cctp/[type].ts`** — calls `mergeCctpBackfill` after the subgraph query; no merge logic in the route itself.
- **`earn/lib/serverPublicClient.ts`** — generalized to a generic `ChainId` so the backfill reuses the existing Origin-allowlisted viem server client instead of a second copy.

## Verification

- **Prod (subgraph-only):** intermittent — 3/8 spaced calls returned the transfer.
- **Local (with backfill):** deterministic — `pending=1` on 20/20 calls; `meta.backfill.addedPending=1` shows it rescuing the subgraph's empty responses live.
- Deposits direction sanity-checked; lint + typecheck clean.

## Notes / follow-ups

- **Per-request cost:** each request now does ~1–2s of RPC (`getLogs` + a few `getBlock`s). Follow-up: a shared cache (one sweep per ~30–60s for all users) to remove per-request latency.
- **`meta.backfill`** is a debug field for observability during the prototype — strip or flag before merge-for-real.
- **Cross-package import:** the backfill (bridge package) imports the earn `serverPublicClient` (app package) — reverse of the usual direction. It resolves because the CCTP route is consumed in the app build; flagging for awareness.
- Stopgap until the subgraph is redeployed/repaired; safe to keep as a standing safety net since it auto-no-ops when the subgraph is healthy.


packages/arb-token-bridge-ui/src/util/cctp/cctpBackfill.ts
```diff
@@ -0,0 +1,332 @@
...
+ * Transient CCTP history backfill (prototype). The subgraph intermittently
+ * returns incomplete `messageSent` results — the decentralized-network gateway
+ * flaps across indexers at different states. This independently re-reads the
+ * recent window from chain logs in the same shape the subgraph returns, so the
+ * API route can merge + dedupe and history is deterministic regardless of which
+ * indexer answered. Additive and fail-open.
+ */
+import { Hex, getAddress, hexToBigInt, hexToNumber, keccak256, parseAbiItem, slice } from 'viem';
+
+import { createServerSidePublicClient } from '@/earn-api/lib/serverPublicClient';
```
fionnachan: i dont think it should come from `@/earn-api/...` so maybe move the utils?

fionnachan (commented):
Overall i prefer the following approach:

1. Use the RPC to get the latest block number of the chain. Then use that block number to query subgraph. This filters out indexers that have not caught up with the latest block number and should solve our intermittent routing to not up-to-date indexer issue. https://thegraph.com/docs/en/subgraphs/querying/distributed-systems/#polling-for-updated-data

2. Query subgraph to get the latest block number the subgraph has synced to (which should be the latest block number of the chain). The metadata returns the block number: https://thegraph.com/docs/en/subgraphs/querying/graphql-api/#subgraph-metadata-example

3. And only try to backfill the data from logs if the block number from the subgraph is not the latest when compared with the RPC.

packages/arb-token-bridge-ui/src/util/rpc/serverPublicClient.ts
```diff
@@ -0,0 +1,44 @@
+import { Chain, PublicClient, createPublicClient, fallback, http } from 'viem';
```
dewanshparashar, line 1: Moved from `earn` folder to here

PR #367, opened 2026-06-22
feat(bridge): add lifi to tx history

chrstph-dvx:
Closes FS-2349


packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/lifi/transactions.ts
```diff
@@ -0,0 +1,228 @@
...
+  return statusResponses
+    .map((statusResponse) => {
+      try {
+        return transformLifiHistoryTransaction({ wallet, statusResponse });
+      } catch {
+        return null;
+      }
+    })
+    .filter((tx): tx is LifiTransactionHistoryItem => tx !== null);
+}
```
chrstph-dvx, line 208: We try catch here, so if one transaction can't be parsed, we still return rest of the transactions

packages/arb-token-bridge-ui/src/components/TransactionHistory/helpers.ts
```diff
@@ -585,53 +586,16 @@ export async function getUpdatedLifiTransfer(
...
-    // Failure
-    if ('timestamp' in statusResponse.sending) {
-      // Source transaction has been executed
-      sourceStatus = WithdrawalStatus.CONFIRMED;
-      destinationStatus = WithdrawalStatus.FAILURE;
-    } else {
-      sourceStatus = WithdrawalStatus.FAILURE;
-      destinationStatus = WithdrawalStatus.UNCONFIRMED;
-    }
+  const { status, destinationStatus, destinationTxId } = getLifiTransferStatus(statusResponse);
```
chrstph-dvx, line 590: No change here, it was extracted to an util, as we need the same logic somewhere else

packages/arb-token-bridge-ui/src/util/normalizeTimestamp.ts
```diff
@@ -0,0 +1,28 @@
+import dayjs from 'dayjs';
```
chrstph-dvx, line 1: This was extracted because it is now imported on the backend, it was causing issues as the previous file it was in was importing client-only files

packages/arb-token-bridge-ui/src/components/TransactionHistory/lifiDisplayUtils.ts
```diff
@@ -0,0 +1,26 @@
...
+  isPending,
+  toAmount,
+}: {
+  isPending: boolean;
+  toAmount: AmountWithToken;
+}) {
+  if (
+    isPending &&
+    toAmount.token.symbol === UNKNOWN_LIFI_TOKEN_SYMBOL &&
+    toAmount.amount.isZero()
```
fionnachan: by this condition, we display the amount even if the symbol is unknown as long as the amount is non-zero? i think that would be strange

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -244,13 +251,112 @@ function getCacheKeyFromTransaction(tx: Transfer) {
...
+    ...apiTx,
+    parentChainId,
+    childChainId,
+    direction: isDepositMode ? 'deposit' : 'withdraw',
+    isWithdrawal: !isDepositMode,
+    resolvedAt: apiTx.resolvedAt ?? localTx.resolvedAt,
+    destinationTxId: apiTx.destinationTxId ?? localTx.destinationTxId,
+    durationMs: apiTx.durationMs || localTx.durationMs,
+    fromAmount: {
+      amount: apiTx.fromAmount.amount || localTx.fromAmount.amount,
```
fionnachan, line 288: `apiTx.fromAmount.amount` is a BigNumber and would always be truthy so we won't get to the fallback amount

fionnachan (dismissed):

packages/arb-token-bridge-ui/src/state/app/utils.ts
```diff
@@ -11,10 +11,12 @@ import {
 import { getUniqueIdOrHashFromEvent } from '../../hooks/useArbTokenBridge';
 import { Transaction } from '../../types/Transactions';
 import { addressesEqual } from '../../util/AddressUtils';
+import { normalizeTimestamp } from '../../util/normalizeTimestamp';
 import { DepositStatus, MergedTransaction } from './state';
 
 export const TX_DATE_FORMAT = 'MMM DD, YYYY';
 export const TX_TIME_FORMAT = 'hh:mm A (z)';
+export { normalizeTimestamp };
```
dewanshparashar: Anti-pattern: Barrel file.. I think it would be much better to only export it from 1 source and fix all downstream imports, even if it inflates the diff by a few no. of files. The resulting code would be much cleaner and easier to use in the future.

dewanshparashar (commented):
Image

not sure if this should be implemented at API level or in `useLifiTransactionHistory()` hook, but we need to filter out same-chain swaps before rendering them in tx history because most likely they are Earn transactions and not bridging.

dewanshparashar (dismissed):

fionnachan (approved):

PR #369, opened 2026-06-23
fix(bridge): use keyed RPC for World Mobile Chain

dewanshparashar:
World Mobile Chain (`869`) shipped Alchemy's throttled `/public` RPC, which caps `eth_getLogs` at 100 blocks and rate-limits hard. With no special handling its transfers never loaded in tx history.

Whitelists WMC in `alchemy.ts` to use our authenticated `/v2/${key}` endpoint (now enabled on our Alchemy app); `mapCustomChainToNetworkData` resolves its RPC through that. Also gets a 10k tx-history block range like the other RaaS chains.

Test for address: `0x0118F25D462Cf3cc9f1f1A7698D8cDB55DBe4850`


fionnachan (commented):
seems that we should update the `orbitChainsData.json` with the API key version alchemy RPC and use it?

In the function as copilot said, we should add a check for the alchemy RPC format

chrstph-dvx (approved):

fionnachan (approved):

PR #371, opened 2026-06-26
feat(bridge): enhance OFT token detection

dewanshparashar:
## Summary

LayerZero OFT tokens must not be moved through the canonical Arbitrum bridge. A canonical deposit lands at the deterministic gateway-derived address on the destination chain, which differs from the token's real OFT deployment — so the funds end up in the wrong token and are effectively stranded. These tokens are meant to move via OFT / LiFi instead.

This PR detects whether a selected token is a LayerZero OFT for the given route and blocks its canonical deposit.

## How it works

- Adds a cached `/api/layerzero-metadata` route that fetches LayerZero's metadata, trims it server-side to just what the client needs (~3.8MB → a small payload), and caches it (data cache + CDN).
- `isBlockedOftDeposit` decides whether a canonical deposit should be blocked:
  - Resolves the deterministic canonical destination address via `getL2ERC20Address` (handles standard, WETH, and custom gateways).
  - Looks up the token's OFT deployment(s) on the destination chain in the metadata and compares:
    - no canonical route (e.g. USDT), or canonical address ≠ the OFT deployment → **block** (route via OFT / LiFi)
    - canonical address == the OFT deployment (adapter over the canonical token, e.g. WETH, USDC.e) → **allow**
  - Falls back to an on-chain `oftVersion()` probe for tokens not yet listed in the metadata (catches native OFTs).

## Tests

- **Unit** — parsing / root-linking logic, malformed-metadata tolerance, and fetch caching + retry (synthetic fixtures).
- **Integration (live)** — validates detection against the real LayerZero metadata endpoint, asserting a known OFT resolves to its real destination deployment.

Closes FS-2352

### DIFF
--- a/packages/arb-token-bridge-ui/src/app/api/layerzero-metadata.ts
+++ b/packages/arb-token-bridge-ui/src/app/api/layerzero-metadata.ts
@@ -5,8 +5,6 @@ import { isRecord } from '../../util/isRecord';
 
 const LAYERZERO_METADATA_URL = 'https://metadata.layerzero-api.com/v1/metadata';
 
-// LayerZero token/chain metadata changes rarely, so cache it for an hour to
-// avoid hammering the upstream API on every client request.
 const CACHE_SECONDS = 60 * 60;
 
 type TrimmedTokenMetadata = {
@@ -15,14 +13,12 @@ type TrimmedTokenMetadata = {
 };
 
 type TrimmedChainMetadata = {
-  chainDetails: { nativeChainId: unknown };
+  chainDetails: { nativeChainId: number };
   tokens: Record<string, TrimmedTokenMetadata>;
 };
 
-// The upstream payload is ~3.8MB; the client only needs each chain's
-// `nativeChainId` and, per token, its `type` and whether it is pegged. Trimming
-// here keeps the cached response small (well under the data cache size limit)
-// and avoids shipping the full payload to every client.
+// The upstream payload is ~3.8MB; keep only what the client needs so the cached
+// response stays under the data cache size limit and isn't shipped in full.
 function trimLayerZeroMetadata(metadata: unknown): Record<string, TrimmedChainMetadata> {
   const trimmed: Record<string, TrimmedChainMetadata> = {};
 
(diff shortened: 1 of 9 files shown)


packages/arb-token-bridge-ui/src/app/api/layerzero-metadata.ts
```diff
@@ -0,0 +1,104 @@
...
+        typeof peggedTo.address === 'string' &&
+        typeof peggedTo.chainName === 'string'
+          ? { peggedTo: { address: peggedTo.address.toLowerCase(), chainName: peggedTo.chainName } }
+          : {};
+    }
+
+    const nativeChainId = Number(chainDetails.nativeChainId);
+
+    trimmed[chainKey] = {
+      chainDetails: { nativeChainId },
```
fionnachan: we shouldn't have chains that we don't care about on the bridge. should be able to filter them out by the chain id here

Image

packages/arb-token-bridge-ui/src/app/api/layerzero-metadata.ts
fionnachan, line 1: goerli, sepolia, and arbitrum sepolia are all present on the LZ API and they are all using the same chain id but this file only picks up the first one

Image

packages/arb-token-bridge-ui/src/util/layerZeroMetadata.ts
```diff
@@ -0,0 +1,139 @@
...
+  childTokenAddresses: string[];
+};
+
+let layerZeroMetadataPromise: Promise<unknown> | null = null;
+
+// The chain + address a token is natively deployed on (its own address if native,
+// else its `peggedTo` target).
+type TokenRoot = LayerZeroPeggedTo;
+
+function getChainByNativeChainId(
```
fionnachan: this is very fragile currently because multiple chains have the same native chain id. For example, Aptos is a Layer-1, and LZ assigns the native chain id `1` to them as well as to Ethereum. We are getting Ethereum correctly now because of Ethereum comes first in the object luckily.

Image

Image

fionnachan (changes requested):

fionnachan (approved):

PR #373, opened 2026-06-30
fix(bridge): search tx history without wallet connection

dewanshparashar:
## What

Transaction history now works for a **searched address** even when no wallet is connected.

## Why

The fetcher short-circuited whenever wagmi's `chain` was undefined, so searching a valid address without a connected wallet returned nothing.

## How

- Extract a `canFetchTransactionHistory` helper that gates fetching on the target address and feature availability — and only requires a connected chain for smart-contract wallets (their history is network-specific).
- Replace `chain.id` reads with a derived `connectedChainId`.
- Add unit tests for the helper.

Closes SUP-1467

### DIFF
--- a/packages/arb-token-bridge-ui/src/hooks/__tests__/canFetchTransactionHistory.test.ts
+++ b/packages/arb-token-bridge-ui/src/hooks/__tests__/canFetchTransactionHistory.test.ts
@@ -1,0 +1,63 @@
+import { describe, expect, it } from 'vitest';
+
+import { canFetchTransactionHistory } from '../canFetchTransactionHistory';
+
+const address = '0x1798440327d78ebb19db0c8999e2368eaed8f413';
+
+describe('canFetchTransactionHistory', () => {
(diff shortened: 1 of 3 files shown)


packages/arb-token-bridge-ui/src/hooks/__tests__/canFetchTransactionHistory.test.ts
```diff
@@ -0,0 +1,63 @@
...
+        address,
+        isLoadingAccountType: false,
+        isTxHistoryEnabled: true,
+        isSmartContractWallet: false,
+        connectedChainId: undefined,
+      }),
+    ).toBe(true);
+  });
+
+  it('blocks fetching for smart contract wallets without a connected chain', () => {
```
fionnachan, line 20: please add a test for smart contract wallet address + connected chain and returning true

currently there's only the falsy case in this file

fionnachan (dismissed):

fionnachan (approved):

PR #375, opened 2026-07-01
feat(bridge): use indexer for canonical txns in tx history

dewanshparashar:
## Summary

Replaces the `experiments=indexer` flag with env-driven per-chain routing so tx history uses the Arbitrum indexer for configured child chains and falls back to subgraph + event logs for the rest.

### Indexer routing & configuration
- Added `NEXT_PUBLIC_INDEXER_CHILD_CHAIN_IDS` env var (documented in `.env.sample`) to list child chain IDs served by the indexer.
- Added `util/txHistory/sources.ts` with `parseChainIds`, `isChildChainIndexed`, and `getCanonicalSource` (`indexer` vs `subgraph`).
- Added unit tests for chain ID parsing and canonical source selection.

### API layer
- Replaced `isIndexerApiExperimentEnabled` with `isIndexerEnabledForRequest`, which routes to the indexer when `l2ChainId` is in the configured list.
- **Deposits**, **withdrawals**, and **eth-deposits-custom-destination** API routes proxy to the indexer for indexed chains.
- **Block number** API (`/api/chains/[chainId]/block-number`) reads from indexer `/status` for indexed chains; subgraph for others. Returns `meta.source: arbitrum-indexer` when applicable.

### Client fetch layer
- Renamed `fetchLatestSubgraphBlockNumber` → `fetchLatestIndexedBlockNumber` in `SubgraphUtils` (works for indexer- or subgraph-backed chains).
- Deposit and withdrawal subgraph fetchers return `[]` instead of throwing when a chain is indexer-backed but lacks a traditional L1/L2 subgraph.
- **Withdrawals**: hybrid fetch — indexed history via API up to the latest indexed block, then event logs for newer blocks. Extracted `fetchWithdrawalsUsingEventLogs` helper.
- Removed the indexer experiment branch that skipped event-log fetching entirely.

### Transaction history UI
- Removed `isExperimentalFeatureEnabled('indexer')` gating from `useTransactionHistory` (deposits, withdrawals, OFT, page sizes).
- Removed the `indexer` entry from experimental feature flags in `util/index.ts`.

## Out of scope / follow-ups
- **Token logos**: `TransactionsTableTokenImage` is left unchanged from `master` to keep this PR scoped to indexer routing. Logo resolution for indexer/testnet/withdrawal txns is tracked in FS-2359; lookup caching in FS-2358.
- **Naming** (`l2ChainId → childChainId`, `fetchWithdrawalsFromSubgraph`): deferred to a codebase-wide `l2/L1 → child/parent` naming sweep.

## Test plan
- [ ] Set `NEXT_PUBLIC_INDEXER_CHILD_CHAIN_IDS` and `INDEXER_API_URL` for an indexed chain
- [ ] Confirm deposits/withdrawals route to indexer API for configured chains
- [ ] Confirm non-configured chains still use subgraph + event logs
- [ ] Verify withdrawal history includes both indexed and recent (event-log) txs
- [ ] Run `pnpm test:ci` (covers `sources.test.ts`)

Closes FS-2348


packages/arb-token-bridge-ui/src/util/withdrawals/fetchWithdrawalsFromSubgraph.ts
```diff
@@ -53,7 +54,7 @@ export async function fetchWithdrawalsFromSubgraph({
   pageNumber?: number;
   searchString?: string;
 }): Promise<WithdrawalFromSubgraph[]> {
-  if (!hasL2Subgraph(Number(l2ChainId))) {
+  if (!hasL2Subgraph(Number(l2ChainId)) && !isChildChainIndexed(Number(l2ChainId))) {
     throw new Error(`L2 subgraph not available for network: ${l2ChainId}`);
   }
```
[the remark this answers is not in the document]
dewanshparashar: Yep - naming will be considered to be changed to an indexer-subgraph agnostic one in the whole codebase in 1 sweep in later PRs.

packages/arb-token-bridge-ui/src/app/api/indexer.ts
```diff
@@ -1,16 +1,17 @@
...
 import { logger } from '../../util/logger';
+import { isChildChainIndexed } from '../../util/txHistory/sources';
 
-export function isIndexerApiExperimentEnabled(request: NextRequest): boolean {
-  const experiments = new URL(request.url).searchParams.get('experiments');
-  return experiments?.split(',').includes('indexer') ?? false;
+export function isIndexerEnabledForRequest(request: NextRequest): boolean {
+  const l2ChainId = new URL(request.url).searchParams.get('l2ChainId');
+
+  return l2ChainId ? isChildChainIndexed(Number(l2ChainId)) : false;
```
fionnachan, line 9: kind of strange that we call it `l2ChainId` instead of `childChainId`?

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
```diff
@@ -5,33 +5,36 @@ import EthereumLogoRoundLight from '@/images/EthereumLogoRoundLight.svg';
...
-function createTokenLogoMapFromTokenLists(tokenLists: TokenListWithId[] | undefined) {
-  if (!tokenLists) {
-    return {};
+function createTokenLogoByAddressMap(tokenLists: TokenListWithId[] | undefined) {
+  const map: { [address: string]: string | undefined } = {};
+  for (const list of tokenLists ?? []) {
+    for (const token of list.tokens) {
+      map[token.address.toLowerCase()] = token.logoURI;
+    }
   }
```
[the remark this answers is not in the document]
fionnachan: agreed with copilot here that the double for loops is bad for performance esp when it can be called by multiple tx history rows

the previous implementation also didn't have cache, please file a ticket to follow up with the performance optimisation when you have time

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
```diff
@@ -5,33 +5,36 @@ import EthereumLogoRoundLight from '@/images/EthereumLogoRoundLight.svg';
...
-  );
+  return map;
 }
 
-export const TransactionsTableTokenImage = ({ tx }: { tx: MergedTransaction }) => {
-  // we need to take token image from mainnet by symbol, some token images don't exists on other networks
-  const tokenLists = useTokenLists(tx.sourceChainId);
-
-  const tokenAddressToLogoSrcMap = createTokenLogoMapFromTokenLists(tokenLists.data);
+function createTokenLogoBySymbolMap(tokenLists: TokenListWithId[] | undefined) {
```
fionnachan: **we shouldn't use this because there could be malicious tokens deployed with the same symbol as an authentic one, like "USDC" but it's a fake USDC token, so we should not display the "USDC" symbol without matching the address

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
```diff
@@ -49,6 +52,13 @@ export const TransactionsTableTokenImage = ({ tx }: { tx: MergedTransaction }) =
     return <Image height={20} width={20} alt="ETH logo" src={EthereumLogoRoundLight} />;
   }
 
+  const logoByAddress = createTokenLogoByAddressMap(sourceChainTokenLists.data);
+  const logoBySymbol = createTokenLogoBySymbolMap(mainnetTokenLists.data);
+
+  const tokenLogoSrc =
+    (tx.tokenAddress ? logoByAddress[tx.tokenAddress.toLowerCase()] : undefined) ??
+    (tx.asset ? logoBySymbol[tx.asset.toLowerCase()] : undefined);
```
fionnachan: i'm worried about fake token display here, esp if another sender sent to the recipient and we display it on the recipient's tx history

packages/arb-token-bridge-ui/src/app/api/indexer.ts
dewanshparashar: Good catch - `l2ChainId` here is the existing query-param key shared across all the tx-history endpoints, and it's forwarded as-is to the indexer/subgraph services, so renaming it is a cross-service contract change (backend would need to accept the new key too) rather than a local rename. I'll fold `l2ChainId → childChainId` into the same codebase-wide `l2/L1 → child/parent` naming sweep, to keep this PR scoped to source routing.

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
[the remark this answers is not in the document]
dewanshparashar: Filed **FS-2358** to add caching as a follow-up (it predates this PR — the address map still carried the original `// TODO: cache`). Note this PR also drops the second (symbol) lookup map, so the per-render work is halved in the meantime.

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
dewanshparashar: Addressed

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
dewanshparashar: Addressed

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
```diff
@@ -49,6 +52,13 @@ export const TransactionsTableTokenImage = ({ tx }: { tx: MergedTransaction }) =
...
+
```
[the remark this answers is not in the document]
dewanshparashar: Fixed in 05677aab9 — alt text now falls back to `'Token logo'` when `tx.asset` is missing.

packages/arb-token-bridge-ui/src/app/api/chains/[chainId]/block-number.ts
```diff
@@ -7,6 +7,31 @@ import {
   getSourceFromSubgraphClient,
 } from '../../../../api-utils/ServerSubgraphUtils';
 import { ChainId } from '../../../../types/ChainId';
+import { isChildChainIndexed } from '../../../../util/txHistory/sources';
+
+type IndexerStatus = Record<string, { id: number; block: { number: number } }>;
```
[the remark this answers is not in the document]
Comment (dewanshparashar) [reply packages/arb-token-bridge-ui/src/app/api/chains/[chainId]/block-number.ts]: Fixed in 05677aab9 — `IndexerStatus.id` is now typed as `string`, matching the indexer serialization (and the test's `String(chainId)`), so the `Number(entry.id)` coercion is meaningful.

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
dewanshparashar: Update: to keep this PR scoped to the indexer change, I reverted `TransactionsTableTokenImage` entirely back to master. The symbol-based fallback (the spoofing risk) was introduced by this PR and is now gone — resolution is address-only, exactly as on master. The proper logo-resolution improvement is queued in FS-2359 (constraint noted there: address-only, never symbol).

fionnachan (approved):

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
dewanshparashar: See above — token-logo file reverted to master, so no symbol-based (spoofable) matching ships in this PR. Follow-up tracked in FS-2359.

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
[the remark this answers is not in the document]
dewanshparashar: Reverted this file to master to keep the PR isolated to the indexer change, so the `alt` handling is now the pre-existing master behavior (not introduced here). Queued the fix in FS-2359.

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableTokenImage.tsx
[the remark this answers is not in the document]
dewanshparashar: Note: this file is now reverted to master (indexer PR kept isolated from tx-history logic), so no change here. Caching still tracked in FS-2358; broader logo-resolution work in FS-2359.

PR #388, opened 2026-07-08
feat(bridge): add chain filtering to tx history

dewanshparashar:
Closes FS-2355

## What

Adds a network filter to the transaction history so you can scope which chains' transactions are fetched and shown, instead of always querying every chain. This shifts RPC load onto the chains actually in use (e.g. a single Orbit chain).

## How it works

- A **"Showing history for:"** dropdown next to the history search — search-by-network, an **All Chains** option, Core/More chain groups with checkboxes, and a testnet toggle.
- A transaction is shown when **either** of its endpoints is a selected chain, so selecting a chain surfaces every route to and from it.
- The RPC fetch list uses the same rule as the display, so the fetched chain pairs and the shown transactions can't disagree. An info tooltip lists exactly which routes are being fetched for the current selection.
- The fetch is debounced, so rapidly toggling several chains coalesces into a single refetch.

## Defaulting

- Until the user touches the filter, it defaults to the **child chain of the pair selected in the bridge** (from the URL query params) — e.g. Arbitrum One for Eth ↔ Arbitrum One. The default is derived at render time rather than written into state, so the very first fetch is already scoped (no unfiltered fetch can fire before the default lands) and changing the bridge pair re-defaults the filter automatically.
- A selection is tagged with the testnet mode it was made in. Switching between testnet and mainnet re-defaults the filter to the new mode's chains; switching back restores the previous selection.
- Transfers initiated during the current session always show, even if the filter doesn't cover their chains, so a just-submitted transaction is never hidden.

## Notes for reviewers

- `util/chainFilter.ts` holds the pure selection logic (`matchesChainFilter`, `resolveSelectedChainIds`) and is unit-tested in isolation.
- `util/txHistoryRoutes.ts` is the single source of truth for route enumeration (canonical via `getMultiChainFetchList`, plus CCTP/OFT/LiFi) — it drives both the filter's chain list and the eligible-routes tooltip.
- CCTP, OFT and LiFi histories are fetched by single API calls, so they are filtered at display time only; the fetch-scoping applies to the per-chain-pair RPC fan-out.

## Tests

- Unit tests for the filter store, the match logic, and default resolution.
- The UI matches the approved Figma design.


packages/arb-token-bridge-ui/src/util/txHistoryRoutes.ts
```diff
@@ -0,0 +1,93 @@
...
+import { ChainId } from '../types/ChainId';
+import { getChains, getChildChainIds, isNetwork } from './networks';
+
+export type ChainPair = { parentChainId: ChainId; childChainId: ChainId };
+
+/**
+ * Every canonical-bridge parent -> child pair across all registered chains: the
+ * list of chain pairs the transaction-history fetcher queries.
+ */
+export function getMultiChainFetchList(): ChainPair[] {
```
chrstph-dvx, line 18: Can we move it from a function to a variable? It should be constant, we don't need to run it on every call

packages/arb-token-bridge-ui/src/components/TransactionHistory/useTransactionHistoryChainFilter.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * The default chain to filter by: the child chain of the pair selected in the
+ * bridge. Combined with either-endpoint matching, this shows every route to and
+ * from that chain — and never comes up empty for a non-canonical pair.
+ *
+ * We read the raw params rather than `useNetworks()` because the latter
+ * sanitizes the pair to a canonical bridge route (e.g. Eth <> Orbit collapses to
+ * Eth <> Arbitrum One). Falls back to the sanitized chains when a param is absent.
+ */
```
chrstph-dvx: `useNetworks` sanitize to valid chain pair. So if we're on `source=Ethereum&destination=ApeChain`, it should sanitize it to Ethereum/ApeChain

packages/arb-token-bridge-ui/src/components/TransactionHistory/useTransactionHistoryChainFilter.ts
```diff
@@ -0,0 +1,53 @@
...
+    () => resolveSelectedChainIds({ selection, isTestnetMode, defaultChainIds }),
+    [selection, isTestnetMode, defaultChainIds],
+  );
+}
+
+// Debounced variant for the (expensive) fetch: toggling several chains in a
+// row coalesces into a single refetch while the checkboxes stay instant.
+export function useDebouncedSelectedChainIds(): number[] {
+  return useDebounce(useSelectedChainIds(), 500);
+}
```
chrstph-dvx: We use this hook in only one place. I would let consumer wrap it with debounce directly

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistoryChainFilter.tsx
```diff
@@ -0,0 +1,288 @@
...
+        sourceChainId: route.parentChainId,
+        destinationChainId: route.childChainId,
+      }),
+    );
+
+    return { filterableChainIds, eligibleRoutes };
+  }, [isTestnetMode, selectedChainIds]);
+}
+
+function CheckboxBox({ checked }: { checked: boolean }) {
```
chrstph-dvx: We already have a Checkbox component (in common/Checkbox), do we need a new design? Can we reuse the existing one for now, and migrate the design later?

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistoryChainFilter.tsx
```diff
@@ -0,0 +1,288 @@
...
+        'flex h-4 w-4 flex-shrink-0 items-center justify-center rounded-sm border-[1px]',
+        checked ? 'border-dark bg-white' : 'border-gray-6 bg-dark',
+      )}
+    >
+      {checked && <CheckIcon className="h-[10px] w-[10px] stroke-[5] text-dark" />}
+    </span>
+  );
+}
+
+function SectionLabel({ children }: { children: React.ReactNode }) {
```
chrstph-dvx: PropsWithChildren

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistoryChainFilter.tsx
```diff
@@ -0,0 +1,288 @@
...
+function EligibleRoutesTooltip({
+  chainPairs,
+  allChainsSelected,
+}: {
+  chainPairs: ChainPair[];
+  allChainsSelected: boolean;
+}) {
+  return (
+    <Tooltip
+      // Focusable trigger so keyboard users can reveal the tooltip.
```
chrstph-dvx, line 95: Awesome, thank you ❤️

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistoryChainFilter.tsx
```diff
@@ -0,0 +1,288 @@
...
+                    </button>
+                  )}
+
+                  {coreChainIds.length > 0 && <SectionLabel>Core Chains</SectionLabel>}
+                  {coreChainIds.map((chainId) => (
+                    <ChainRow
+                      key={chainId}
+                      chainId={chainId}
+                      checked={isChainChecked(chainId)}
+                      onClick={() => toggleChain(chainId)}
```
chrstph-dvx: Same on L268. We can pass toggleChain only instead of an anonymous function, otherwise it's a new ref every time. ChainRow has access to chainId, so it can call toggleChain(chainId) itself

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -571,6 +554,11 @@ const useTransactionHistoryWithoutStatuses = (address: Address | undefined) => {
 
   const forceFetchReceived = useForceFetchReceived((state) => state.forceFetchReceived);
 
+  // Debounced so rapidly toggling several chains coalesces into a single refetch.
+  const selectedChainIds = useDebouncedSelectedChainIds();
+  // Stable identifier for the selected chains so SWR refetches when the filter changes.
+  const selectedChainIdsKey = [...selectedChainIds].sort((a, b) => a - b).join(',');
```
chrstph-dvx, line 567: NIT: Do we need the spread here? useMemo return a new array per caller, it's not shared across all callers

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -758,13 +768,25 @@ const useTransactionHistoryWithoutStatuses = (address: Address | undefined) => {
...
 
+  // CCTP, OFT and LiFi are fetched by single API calls rather than the scoped
+  // chain-pair fan-out, so filter them here — before pagination — to keep
+  // unrelated transfers from consuming page slots and per-tx mapping work.
+  const matchesSelectedChains = (tx: MergedTransaction) =>
+    matchesChainFilter({
+      selectedChainIds,
+      sourceChainId: tx.sourceChainId,
+      destinationChainId: tx.destinationChainId,
+    });
```
chrstph-dvx: This can be moved outside of this function

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -758,13 +768,25 @@ const useTransactionHistoryWithoutStatuses = (address: Address | undefined) => {
...
     ...deposits,
     ...withdrawals,
-    ...(isTestnetMode ? combinedCctpTestnetTransfers : combinedCctpMainnetTransfers),
-    ...oftTransfers,
-    ...lifiHistoryTransactions,
+    ...(isTestnetMode ? combinedCctpTestnetTransfers : combinedCctpMainnetTransfers).filter(
+      matchesSelectedChains,
+    ),
+    ...oftTransfers.filter(matchesSelectedChains),
+    ...lifiHistoryTransactions.filter(matchesSelectedChains),
```
chrstph-dvx, line 791: Instead of fetching all transferring, and filtering on the frontend, can we directly filter everything on the backend?

packages/arb-token-bridge-ui/src/util/txHistoryRoutes.test.ts
```diff
@@ -0,0 +1,49 @@
...
+describe('getTxHistoryRoutes', () => {
+  it('returns routes for both network modes', () => {
+    expect(getTxHistoryRoutes({ isTestnetMode: false }).length).toBeGreaterThan(0);
+    expect(getTxHistoryRoutes({ isTestnetMode: true }).length).toBeGreaterThan(0);
+  });
+
+  it('only returns routes whose parent endpoint matches the requested mode', () => {
+    // Child endpoints aren't asserted: some LiFi destinations (e.g. ApeChain,
+    // Superposition) are registered at runtime, so `isNetwork` misreports them
+    // in the unit-test environment.
```
chrstph-dvx: We can register orbit chain for tests, we have this pattern in few places:
`registerCustomArbitrumNetwork(orbitChainsData.mainnet.find((chain) => chain.chainId === ChainId.ApeChain)!);
`

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
dewanshparashar: Can do in a follow up PR. But for this PR's scope of cutting down on RPC load, these external vendor queries are not load bearing. Currently these APIs return the data in 1 go without any filters.

chrstph-dvx (approved):

PR #390, opened 2026-07-08
fix(bridge): keep refunded tx in tx history

chrstph-dvx:
### Steps to test

image

### DIFF
--- a/packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/lifi/transactions.ts
+++ b/packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/lifi/transactions.ts
@@ -106,13 +106,14 @@ export function transformLifiHistoryTransaction({
     receiving ? receiving.chainId || lastStep?.toToken.chainId : lastStep?.toToken.chainId,
   );
   const isPending = statusResponse.status === 'PENDING';
+  const isRefunded = statusResponse.status === 'DONE' && statusResponse.substatus === 'REFUNDED';
 
   if (!sourceChainId || !destinationChainId) {
     return null;
   }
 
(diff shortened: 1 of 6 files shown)


packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/lifi/transactions.ts
```diff
@@ -106,13 +106,14 @@ export function transformLifiHistoryTransaction({
...
   const isPending = statusResponse.status === 'PENDING';
+  const isRefunded = statusResponse.status === 'DONE' && statusResponse.substatus === 'REFUNDED';
 
   if (!sourceChainId || !destinationChainId) {
     return null;
   }
 
   // Ignore earn transactions (same chain swap)
-  if (sourceChainId === destinationChainId) {
+  if (sourceChainId === destinationChainId && !isRefunded) {
```
dewanshparashar, line 116: The fix will truly be complete when you update the earn integrator to something different from the portal. Only then the list being shown in tx history will contain bridge-originated txns.

dewanshparashar (commented):
integrator id needs to be taken care of for the sake of completeness

packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/lifi/transactions.ts
chrstph-dvx: Bridge doesn't have same chain swaps, so it exclude all earn transactions, regardless of the integrator id

dewanshparashar (dismissed):

dewanshparashar (approved):

PR #403, opened 2026-07-14
fix(bridge): fix race condition in transferDisabledDialog

chrstph-dvx:

### DIFF
--- a/packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanel.eth-weth.integration.test.tsx
+++ b/packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanel.eth-weth.integration.test.tsx
@@ -16,6 +16,7 @@ import {
   tokenExpectationsByChain,
   wethTokenExpectation,
 } from './TransferPanel.integration.helpers';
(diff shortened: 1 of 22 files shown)


packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanel.eth-weth.integration.test.tsx
```diff
@@ -135,8 +209,31 @@ async function assertEthWethSelection({
...
+    mockedOftBlock.enabled = false;
+    mockedTokenLists.unresolved = false;
+  });
+
   it.each(ethWethCases)(
     'renders expected source and destination tokens for ETH/WETH override: $sourceChain -> $destinationChain',
     assertEthWethSelection,
   );
+
+  it('does not show the disabled-transfer dialog while LiFi token lists are unresolved', async () => {
```
dewanshparashar: Can we also add a test for the case where the transfer disabled dialog is expected to show? just to make sure this is not a false positive.

dewanshparashar (dismissed):

dewanshparashar (dismissed):

dewanshparashar (approved):

PR #405, opened 2026-07-15
feat: upgrade pnpm to 11.13.0

dewanshparashar:
### Summary

Upgrades pnpm from `10.15.1` to `11.13.0`. pnpm 11 uses npm's bulk advisory endpoint, so `pnpm audit` / `audit:ci` work again after npm retired the legacy endpoint (HTTP 410). This is the repo-wide alternative to the CI-only workaround in #404 — pick one.

pnpm 11 no longer reads settings from `package.json`/`.npmrc`, so they move into `pnpm-workspace.yaml`:

- `pnpm.overrides` → `overrides`
- `.npmrc` (`shamefully-hoist`, `auto-install-peers`, `dedupe-peer-dependents`, `resolve-peers-from-workspace-root`) → equivalent keys; `.npmrc` removed (it only held these)
- `onlyBuiltDependencies` → `allowBuilds` (renamed in v11)
- `trustPolicyExclude` added for 13 well-known packages (semver, @noble/\*, react-redux, …) that v11's `no-downgrade` trust policy flags because attestation was added only in newer versions — the pinned versions are unchanged and legitimate

`pnpm-lock.yaml` is unchanged — dependency versions are identical.

### Steps to test

- `pnpm install` builds native modules cleanly; `pnpm lint` passes (tsc clean); `pnpm audit:ci` → "Passed pnpm security audit".
- CI green.


pnpm-workspace.yaml
```diff
@@ -30,3 +87,17 @@ minimumReleaseAge: 1440
 # Prevent installing package versions with decreased trust levels
 # compared to previously published versions.
 trustPolicy: no-downgrade
+trustPolicyExclude:
+  - '@coinbase/cdp-sdk@1.38.4'
```
dewanshparashar, line 95: Need to check in depth what we can do about this field.

chrstph-dvx (approved):

PR #293, opened 2026-04-21
feat(bridge): migrate rainbowkit to reown

chrstph-dvx:

PR #407, opened 2026-07-16
fix: pass import+approve token e2e tests [PoC]

dewanshparashar:
Important: this is just a POC. It is definitely worth breaking this into 3-4 focused, production-ready PRs. It is marked ready-for-review to let E2Es run.

The import token E2E tests were failing on every PR for months. It turned out to be five separate issues stacked on top of each other. Fixing all five (one commit each) got the tests passing.

## What was wrong, step by step

1. **The E2E wagmi chain list was missing Arbitrum One.** The import test bridges Ethereum -> Arbitrum One, and since the reown migration (#293) a destination chain missing from the AppKit config breaks the page. Added `arbitrum` to the E2E chain list.

2. **The lifi token registry was too big to cache.** The serialized registry (~2.6MB) is over Next.js' 2MB `unstable_cache` limit, so every cache write silently failed and every request refetched everything from li.quest. Trimmed the cached value to ~170KB (route output verified identical for all 25 chain pairs).

3. **The token search store was derived in a second async hook**, so it could lag behind the fetched lists. It's a cheap synchronous derivation, so it now runs in `useMemo` during render.

4. **The E2E login helper passed an invalid chain pair.** It forced `destinationChain=nitro-testnode-l3` even when a test explicitly connects to Ethereum. It now only applies the L3 default when no network override is given.

5. **SWR was losing the token lists result.** CI instrumentation showed the fetch resolving while every subscriber kept getting `data: undefined, isLoading: false` — indefinitely, with React rendering normally. `useTokenLists` now uses a small module-level store (cache + in-flight dedupe + `useSyncExternalStore`) instead of SWR, so a subscriber can never miss the result.

## Result

All 5 import E2E jobs green: https://github.com/OffchainLabs/arbitrum-portal/actions/runs/29520381584


packages/arb-token-bridge-ui/src/components/TransferPanel/TokenSearchUtils.ts
```diff
@@ -17,17 +17,21 @@ export function useTokensFromLists() {
...
-      tokenListsToSearchableTokenStorage(
-        _tokenLists,
-        String(_parentChainId),
-        String(_childChainId),
-      ),
-  );
-
-  return { data, isLoading: isLoadingTokenLists || isLoading };
+  // Derived synchronously so the tokens can never lag behind the fetched lists,
+  // which an extra async hook allowed under heavy re-rendering
```
chrstph-dvx, line 21: we used SWR because otherwise each caller of this hook would fetch and return their own instances, this was causing performance issues

dewanshparashar: Understood. Worth noting that this is only a PoC which tries to get E2Es passing again by fixing some erroring parts of the bridge. If removing token-list SWR got the E2Es passing, then it doesn't necessarily mean we will remove SWR, but that the current implementation is worth going over and optimizing a bit to find out what's wrong with it. Just gives us a starting point of where to poke.

PR #419, opened 2026-07-22
ci(monitoring): add enterprise chain slack alerts

dewanshparashar:
## Summary

Monitoring alerts for enterprise chains (currently Robinhood) now fire in their own Slack channel instead of the shared orbit channels.

- `networks.ts`: `enterpriseChainIds` + `isEnterpriseChain()` — product-wide classification
- `generateOrbitChainsToMonitor`: `MONITOR_ENTERPRISE_CHAINS` flag picks enterprise chains or all other orbit chains
- `monitor-config.json` + monitor workflows: new `enterprise` group using `ENTERPRISE_CHAIN_MONITORING_SLACK_TOKEN` / `_CHANNEL`
- Unit tests for the chain selection

No changes needed in arbitrum-monitoring.

## Before merge

- [x] Add the `ENTERPRISE_CHAIN_MONITORING_SLACK_TOKEN` / `_CHANNEL` repo secrets.

### DIFF
--- a/.github/workflows/assertion-monitor.yml
+++ b/.github/workflows/assertion-monitor.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   run-monitoring:
-    name: Assertion Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
+    name: Assertion Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain == 'enterprise' && 'Enterprise' || matrix.chain }})
     strategy:
       matrix:
         chain: [core, orbit, enterprise]
--- a/.github/workflows/assertion-monitor.yml
+++ b/.github/workflows/assertion-monitor.yml
@@ -10,7 +10,7 @@ jobs:
     name: Assertion Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
     strategy:
       matrix:
-        chain: [core, orbit]
+        chain: [core, orbit, enterprise]
     uses: ./.github/workflows/monitoring.yml
     with:
       chain: ${{ matrix.chain }}
--- a/.github/workflows/batch-poster-monitor.yml
+++ b/.github/workflows/batch-poster-monitor.yml
@@ -10,7 +10,7 @@ jobs:
     name: Batch Poster Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
     strategy:
       matrix:
-        chain: [core, orbit]
+        chain: [core, orbit, enterprise]
     uses: ./.github/workflows/monitoring.yml
     with:
       chain: ${{ matrix.chain }}
--- a/.github/workflows/batch-poster-monitor.yml
+++ b/.github/workflows/batch-poster-monitor.yml
@@ -7,7 +7,7 @@ on:
 
 jobs:
   run-monitoring:
-    name: Batch Poster Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
+    name: Batch Poster Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain == 'enterprise' && 'Enterprise' || matrix.chain }})
     strategy:
       matrix:
         chain: [core, orbit, enterprise]
--- a/.github/workflows/monitor-config.json
+++ b/.github/workflows/monitor-config.json
@@ -16,6 +16,7 @@
   "orbit": {
     "generateCommand": "generateOrbitChainsToMonitor",
     "configFile": "__auto-generated-orbit-chains.json",
+    "excludeChainIds": [4663],
     "slackTokens": {
       "assertion": "ORBIT_CHAIN_ASSERTION_MONITORING_SLACK_TOKEN",
       "batch-poster": "ORBIT_CHAIN_BATCH_POSTER_MONITORING_SLACK_TOKEN",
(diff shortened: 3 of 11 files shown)


.github/workflows/retryable-monitor.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     name: Retryable Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
     strategy:
       matrix:
-        chain: [core, orbit]
+        chain: [core, orbit, enterprise]
```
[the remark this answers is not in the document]
dewanshparashar: Fixed in 6d663ee4f

.github/workflows/batch-poster-monitor.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     name: Batch Poster Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
     strategy:
       matrix:
-        chain: [core, orbit]
+        chain: [core, orbit, enterprise]
```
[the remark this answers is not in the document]
dewanshparashar: Fixed in 6d663ee4f

.github/workflows/assertion-monitor.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     name: Assertion Monitor (${{ matrix.chain == 'orbit' && 'Orbit' || matrix.chain == 'core' && 'Core' || matrix.chain }})
     strategy:
       matrix:
-        chain: [core, orbit]
+        chain: [core, orbit, enterprise]
```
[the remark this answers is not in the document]
dewanshparashar: Fixed in 6d663ee4f

packages/arb-token-bridge-ui/scripts/utils.test.ts
```diff
@@ -0,0 +1,38 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+
+import { ChainId } from '../src/types/ChainId';
+import { isEnterpriseChain } from '../src/util/networks';
+import { getOrbitChainsOutputFile, getOrbitChainsToMonitor } from './utils';
+
+describe('getOrbitChainsToMonitor', () => {
+  afterEach(() => {
+    vi.unstubAllEnvs();
+  });
```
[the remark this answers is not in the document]
dewanshparashar: isE2eTestingEnvironment is a module-level const evaluated at import time, so stubbing the env var inside the test body wouldn't have any effect. The flag isn't set in the vitest env — and if it were, existing suites like networks.test.ts would break the same way, so this test makes no new assumption.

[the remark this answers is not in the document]
dewanshparashar: the flag isn't set in the vitest environment.

packages/arb-token-bridge-ui/scripts/utils.ts
```diff
@@ -49,3 +49,23 @@ export const getChainToMonitor = ({
...
+  const orbitChains = getOrbitChains({ mainnet: true, testnet: false });
+
+  if (isEnterpriseMode()) {
+    return orbitChains.filter((orbitChain) => isEnterpriseChain(orbitChain.chainId));
+  }
+
+  return orbitChains.filter((orbitChain) => !isEnterpriseChain(orbitChain.chainId));
+}
+
+export function getOrbitChainsOutputFile() {
```
douglance, line 65: two sources of truth for the filename: decided here off MONITOR_ENTERPRISE_CHAINS, but the Copy step in monitoring.yml reads it from configFile. if they drift the copy fails with file-not-found. cross-ref or derive one from the other.

packages/arb-token-bridge-ui/src/util/networks.ts
```diff
@@ -559,6 +559,12 @@ export function isCoreChainForDisplay(chainId: number) {
   return isNetwork(chainId).isCoreChain || uiCoreChainIds.includes(chainId);
 }
 
+const enterpriseChainIds: number[] = [ChainId.RobinhoodChain];
```
douglance, line 563: RobinhoodChain is also in uiCoreChainIds above (display-core), so worth a comment that this list is monitoring-routing-only. also: if it ever empties, the enterprise leg still runs and monitors nothing.

.github/workflows/monitor-config.json
```diff
@@ -26,5 +26,19 @@
       "batch-poster": "ORBIT_CHAIN_BATCH_POSTER_MONITORING_SLACK_CHANNEL",
       "retryable": "ORBIT_RETRYABLE_MONITORING_SLACK_CHANNEL"
     }
+  },
+  "enterprise": {
+    "generateCommand": "generateOrbitChainsToMonitor",
+    "configFile": "__auto-generated-enterprise-chains.json",
+    "slackTokens": {
+      "assertion": "ENTERPRISE_CHAIN_MONITORING_SLACK_TOKEN",
```
douglance, line 34: all 3 monitor types share one token/channel here — fine (single channel), just means per-type routing later = editing 3 identical lines.

packages/arb-token-bridge-ui/scripts/utils.test.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    expect(chainIds.filter(isEnterpriseChain)).toEqual([]);
+  });
+
+  it('returns only enterprise chains in enterprise mode', () => {
+    vi.stubEnv('MONITOR_ENTERPRISE_CHAINS', 'true');
+
+    const chainIds = getOrbitChainsToMonitor().map((orbitChain) => orbitChain.chainId);
+
+    expect(chainIds).toEqual([ChainId.RobinhoodChain]);
```
douglance: exact-match, so this breaks when a 2nd enterprise chain is added. fine as a guard, just flagging.

douglance (commented):
clean approach, one predicate driving both legs and core is hardcoded so robinhood cant double-monitor. logic looks right.

one non-code thing before merge: this moves robinhood off the orbit leg to enterprise-only, which reads ENTERPRISE_CHAIN_MONITORING_SLACK_TOKEN/_CHANNEL via secrets: inherit. if those arent set, robinhood alerts silently go nowhere. can we confirm they exist? (run 29942646332 is smoke-testing this.)

packages/arb-token-bridge-ui/scripts/utils.test.ts
dewanshparashar: Fixed in 173230e00 — assertions now derive from isEnterpriseChain instead of exact-matching, so a 2nd enterprise chain won't break it.

packages/arb-token-bridge-ui/src/util/networks.ts
dewanshparashar: Added a one-liner in 173230e00. On the empty-list case: the enterprise leg would generate an empty childChains and the monitors no-op — accepting that as a harmless idle run rather than adding a workflow guard.

packages/arb-token-bridge-ui/scripts/utils.ts
dewanshparashar: Acknowledged, keeping as-is: drift fails loudly at the copy step (file-not-found) rather than misrouting anything, and both names are defined side by side in this PR. Deriving one from the other would mean passing the filename through env again, which we intentionally simplified away in favor of the single flag.

fionnachan (approved):

PR #422, opened 2026-07-23
feat(bridge): serve CCTP history from the indexer replica

douglance:
CCTP history queries now go to the arbitrum-indexer's drop-in replica of Circle's CCTP v1 subgraphs, falling back to the Circle subgraph on any query error.

The indexer base URL is resolved per chain from `INDEXER_API_URL_BY_CHAIN` — a JSON object keyed by chain ID, e.g. `{"42161":"https://indexer.example"}` — because chains are split across separate indexer deployments rather than served by one host. Each chain's queries go to `<its URL>/api/v1/cctp/graphql/:chainId`.

A chain missing from the map — or an unparseable map, or a non-URL value — resolves to "no indexer", and that chain's queries go straight to the subgraph. Behavior is unchanged wherever the variable isn't configured.

**Deploy note:** `INDEXER_API_URL_BY_CHAIN` is set on this project for production, preview and development, so no action is needed at merge. Until a chain is present in the map, its CCTP history comes from the Circle subgraphs.

The other `INDEXER_API_URL` consumers (bridge history proxy, block-number route) still read the single-host variable and migrate separately. When they do, they must resolve by child chain ID — the parent-chain entries point at the deployment that serves the CCTP replica, which is not necessarily the one indexing a given child chain.

Verified on the preview deployment: `meta.source` reports the indexer for mainnet and testnet, deposits and withdrawals; transfer sets and every non-`id` field match what the Circle subgraphs return for the same wallets.

### DIFF
--- a/packages/arb-token-bridge-ui/src/api-utils/ServerSubgraphUtils.ts
+++ b/packages/arb-token-bridge-ui/src/api-utils/ServerSubgraphUtils.ts
@@ -1,4 +1,13 @@
-import { ApolloClient, HttpLink, InMemoryCache, NormalizedCacheObject } from '@apollo/client';
+import {
+  ApolloClient,
+  ApolloQueryResult,
+  HttpLink,
+  InMemoryCache,
+  MaybeMasked,
+  NormalizedCacheObject,
+  OperationVariables,
+  QueryOptions,
+} from '@apollo/client';
 import ApolloLinkTimeout from 'apollo-link-timeout';
 
 import { ChainId } from '../types/ChainId';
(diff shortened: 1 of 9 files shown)


packages/arb-token-bridge-ui/src/api-utils/ServerSubgraphUtils.ts
```diff
@@ -213,11 +344,5 @@ export function getL2SubgraphClient(l2ChainId: number) {
...
   subgraphClient: ApolloClient<NormalizedCacheObject>,
 ): string | null {
-  const uri = (subgraphClient.link as any).options?.uri;
-
-  if (typeof uri === 'undefined') {
-    return null;
-  }
-
-  return new URL(uri).origin;
+  return subgraphClientSources.get(subgraphClient) ?? null;
```
dewanshparashar: I think the source should not reveal our indexer endpoint (uri) since have kept it hidden behind the API. We should reveal the source-key like "arbitrum-indexer" or "cctp-subgraph" and nothing more.

packages/arb-token-bridge-ui/src/api-utils/ServerSubgraphUtils.ts
```diff
@@ -159,23 +172,141 @@ function createSubgraphClient(key: SubgraphKey) {
...
+  return {
+    query: async <T = unknown, TVariables extends OperationVariables = OperationVariables>(
+      options: QueryOptions<TVariables, T>,
+    ): Promise<CctpQueryResult<T>> => {
+      try {
+        return {
+          ...(await indexerClient.query<T, TVariables>(options)),
+          source: indexerSource,
+        };
+      } catch (error) {
```
dewanshparashar: I think if the indexer URL is configured for a chain, we should only query the indexer and not silently fall back to the subgraph. With a silent fallback we won't promptly know when our indexer infra is shaky (we got burned by exactly this when the CCTP subgraph indexer froze and nobody noticed for weeks). The env var is already our kill switch: if the indexer misbehaves we remove that chain from the map and we're back on the subgraph in one deploy. A failed indexer query should just error the request and report to Sentry so it stays visible.

This would also simplify the PR a lot: both branches of getCctpSubgraphClient return a plain ApolloClient, so the fallback wrapper, the lazy client, the custom client type and most of the source plumbing go away, and the route barely changes.

dewanshparashar (commented):
I have a feeling this PR is a bit more complex than what the feature asks for. The feature is basically: if an indexer URL is configured for the chain, use the indexer, otherwise use the subgraph. Most of the extra machinery here (the WeakMap source registry, the custom CctpSubgraphClient type, the per-query source on results) exists only to make meta.source report the exact origin per query, and even then it's only accurate for the messageSents query, not the messageReceiveds ones. Combined with my comments about the fallback and the source key, I think this can collapse into a small env-driven routing change.

packages/arb-token-bridge-ui/src/api-utils/ServerSubgraphUtils.ts
```diff
@@ -159,23 +172,141 @@ function createSubgraphClient(key: SubgraphKey) {
...
+ *
+ * Returns `undefined` for anything unusable — unparseable JSON, a chain that
+ * isn't in the map, a value that isn't a URL — so a misconfigured map degrades
+ * to the subgraph instead of failing the request.
+ */
+function getIndexerApiUrl(chainId: number): string | undefined {
+  let urlByChainId: Record<string, unknown>;
+
+  try {
+    const parsed: unknown = JSON.parse(process.env.INDEXER_API_URL_BY_CHAIN || '{}');
```
dewanshparashar: Can we also add INDEXER_API_URL_BY_CHAIN to .env.sample? The existing indexer vars are documented there. Maybe with a small note that this is intentionally separate from INDEXER_API_URL (which serves canonical tx history for RH), so nobody consolidates the two later and breaks that path.

packages/arb-token-bridge-ui/src/api-utils/ServerSubgraphUtils.ts
dewanshparashar: Bumping this since the stack ships incrementally: this PR is the one that starts putting the indexer origin in meta.source in production, so the fix should land here and not wait for #469. A static key like `arbitrum-indexer` on the indexer path and the subgraph key on the fallback path would do it.

douglance: Done in e986d879 — labels only, no URL can reach the response now.

The label is chosen where the backend is chosen rather than derived from the URI: `arbitrum-indexer` on the indexer path, and the `SubgraphKey` (`cctp-ethereum`, `l1-arbitrum-one`, …) on the subgraph and fallback paths. `createApolloClient` no longer touches the source map at all — it only builds URLs, so it was never in a position to name the source. The `WeakMap` value type is narrowed to `SubgraphSource`, which is the part that keeps a URL from being stored again.

Two things worth flagging since the diff does not show them:

- **It relabels the non-CCTP routes too, with zero lines changed in them.** `deposits`, `withdrawals`, `eth-deposits-custom-destination` and `block-number`'s subgraph branch all build clients through `createSubgraphClient`, so their `meta.source` goes from a gateway origin to e.g. `l1-arbitrum-one`. Same leak class, and it settles `block-number`'s own split where line 80 emitted a label and line 111 an origin in the same response type. Nothing in the repo reads `meta.source` — every fetch helper destructures only `.data` — so there are no consumer changes.
- **What we lose:** the origin used to say *which* Graph path served a query (pinned-indexer vs Graph Network vs self-hosted) for the same key, which is relevant to the graph-node #6683 pinning. `createSubgraphClient` already `logger.debug`s the deployment and pinned indexer, so it stays in server logs. I did not invent extra vocabulary like `cctp-ethereum-pinned` — you said nothing more.

Left alone deliberately: `INDEXER_SOURCE` stays module-private even though `block-number.ts:80` hardcodes the same literal. That duplication predates this PR, and importing the constant there would need `block-number.test.ts:15`'s `vi.mock` factory extended or the binding resolves to `undefined` — so the literal ends up written twice regardless. Better as a follow-up in #469, which touches that plumbing anyway.

New test pins the invariant rather than the vocabulary: it asserts the reported source is not `URL.canParse`-able and contains neither the indexer host nor the API key, across both the fallback and recovered queries. Verified it fails against the previous code (`expected true to be false`).

dewanshparashar (dismissed):

dewanshparashar (dismissed):

dewanshparashar (approved):

PR #423, opened 2026-07-24
fix(bridge): hide zero native gas costs

chrstph-dvx:
## Summary
- omit exact-zero gas rows from native bridge route estimates
- keep non-zero parent and child gas currencies displayed independently
- avoid rendering a combined gas row when its total is zero

## Testing
ArbitrumOne to ApeChain: Native route shows "X ETH and 0 APE" on production, it should shows "X ETH" only in this branch

### DIFF
--- a/packages/arb-token-bridge-ui/src/components/TransferPanel/Routes/getGasCostAndToken.test.ts
+++ b/packages/arb-token-bridge-ui/src/components/TransferPanel/Routes/getGasCostAndToken.test.ts
@@ -162,6 +162,22 @@ describe('getGasCostAndToken', () => {
         });
       },
     );
+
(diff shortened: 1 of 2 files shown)


packages/arb-token-bridge-ui/src/components/TransferPanel/Routes/getGasCostAndToken.test.ts
```diff
@@ -185,44 +201,6 @@ describe('getGasCostAndToken', () => {
           ],
         },
       },
-      {
```
chrstph-dvx, line 188: Those tests were duplicate (same parameters)

fionnachan (approved):

PR #427, opened 2026-07-27
fix(bridge): allow import of tokens on robinhood

chrstph-dvx:


packages/arb-token-bridge-ui/src/util/TokenListUtils.ts
```diff
@@ -16,6 +17,20 @@ const UniswapLogo = '/images/lists/uniswap.png';
...
+export function isTokenAvailableOnChain(
+  token: ERC20BridgeToken | undefined,
+  chainId: number,
+): boolean {
+  if (token?.chainId === undefined) {
+    return true;
+  }
+
+  const isSingleChainToken =
+    token.l2Address === undefined || addressesEqual(token.address, token.l2Address);
```
fionnachan: why does `addressesEqual(token.address, token.l2Address)` mean the token is only on a single chain?

packages/arb-token-bridge-ui/src/components/TransferPanel/TokenImportDialog.tsx
```diff
@@ -71,10 +77,19 @@ export function TokenImportDialog({
...
-  const { data: l1Address, isLoading: isL1AddressLoading } = useERC20L1Address({
+  const {
+    data: l1Address,
+    hasParentAddress,
+    isLoading: isL1AddressLoading,
+  } = useERC20L1Address({
     eitherL1OrL2Address: tokenAddress,
     l2Provider: childChainProvider,
   });
+  const addressToImport = l1Address ?? standaloneTokenAddress;
```
fionnachan: this `standaloneTokenAddress` fallback caused L208 `if (!isL1AddressLoading && !addressToImport) {` to never be true

packages/arb-token-bridge-ui/src/hooks/useSelectedToken.ts
fionnachan, line 1: please add some new tests to `useSelectedToken.test.ts` to cover the changes

packages/arb-token-bridge-ui/src/hooks/useSelectedToken.ts
```diff
@@ -100,23 +101,36 @@ export const useSelectedToken = (): [
...
           if (sanitizedTokenAddress) {
+            const tokenAddress = erc20ParentAddress?.toLowerCase() ?? '';
+            const token =
+              tokenOverride || tokensFromUser[tokenAddress] || tokensFromLists[tokenAddress];
             return {
               token: sanitizedTokenAddress,
-              destinationToken: sanitizedTokenAddress,
+              destinationToken: isTokenAvailableOnChain(token, networks.destinationChain.id)
+                ? sanitizedTokenAddress
+                : undefined,
```
fionnachan: seems unexpected to me that `destinationToken` can be `undefined` given that it'd either be a swap or a transfer?

before this branch, it was either `null` or zero address for native gas token?

packages/arb-token-bridge-ui/src/components/TransferPanel/TokenSearch.tsx
```diff
@@ -430,9 +436,19 @@ function TokensPanel({
...
-    if (networks.sourceChain.id === ChainId.RobinhoodChain) {
-      setErrorMessage(error);
-      return;
+    if (lifiChainIdsAllowingUnmatchedTokens.includes(networks.sourceChain.id)) {
+      try {
+        await token.add(newToken);
+        return;
+      } catch {
+        try {
+          await token.addStandaloneToken(newToken, networks.sourceChain.id);
```
fionnachan: there are two ENAs on the destination token search panel, can you please check what's up? looks like it could be because of `registry.ts CUSTOM_TOKENS`? (only on Robinhood Chain -> Arbitrum One) https://arbitrum-portal-git-allow-robinhood-token-import-offchain-labs.vercel.app/bridge?amount=1&destinationChain=arbitrum-one&sanitized=true&sourceChain=robinhood-chain

Image

packages/arb-token-bridge-ui/src/components/TransferPanel/DestinationTokenSearch.tsx
fionnachan, line 1: while testing, i notice that this is the behaviour already on prod, but we should change it

we should allow selection of "USDC" on Ethereum (destination chain) for swaps but currently users cannot select USDC.
Image

packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanel.robinhood-token-import.integration.test.tsx
```diff
@@ -0,0 +1,357 @@
...
+});
+
+vi.mock('../../hooks/useERC20L1Address', () => ({
+  useERC20L1Address: ({ eitherL1OrL2Address }: { eitherL1OrL2Address: string }) => ({
+    data:
+      eitherL1OrL2Address.toLowerCase() === ROBINHOOD_BWLK_ADDRESS ? ETHEREUM_BWLK_ADDRESS : null,
+    hasParentAddress: eitherL1OrL2Address.toLowerCase() === ROBINHOOD_BWLK_ADDRESS,
+    isLoading: false,
+  }),
+}));
```
fionnachan: i don't think you need this mock because `getL1ERC20Address` is already mocked on this file, otherwise the `getL1ERC20Address` mock is not used

packages/arb-token-bridge-ui/src/components/TransferPanel/TokenSearch.tsx
chrstph-dvx: LiFi returns two distinct ENA contracts on Arbitrum One, 0x585... and 0xdf8..., and both map to the same Robinhood ENA contract. The selections were already setting different destinationToken values, but the rows incorrectly displayed the Robinhood child-chain address for both.

I updated destination rows to display the contract on the actual destination chain, so the two distinct Arbitrum addresses are now visible.

packages/arb-token-bridge-ui/src/components/TransferPanel/DestinationTokenSearch.tsx
chrstph-dvx: It's a bigger refactor outside of the scope of this PR. We would need to allow tokens on the destination chain without counterpart on the source chain. This would enable >5k tokens on Ethereum, the bridge currently can't handle it.

packages/arb-token-bridge-ui/src/components/TransferPanel/TransferPanel.robinhood-token-import.integration.test.tsx
chrstph-dvx: Good catch

packages/arb-token-bridge-ui/src/hooks/useSelectedToken.ts
chrstph-dvx: This is not introduced by this branch. destinationToken in query param was always `string | undefined`. You could pass `null` to `setSelectedToken`, but you could not set `destinationToken` to null.

`useDestinationToken` will resolve to the proper token here

AGENTS.md
fionnachan, line 1: i don't think this should be committed as it's not a shared tool, at least for now

packages/arb-token-bridge-ui/src/hooks/useArbTokenBridge.ts
```diff
@@ -344,6 +295,45 @@ export const useArbTokenBridge = (params: TokenBridgeParams): ArbTokenBridge =>
...
+        name,
+        type: TokenType.ERC20,
+        symbol,
+        address,
+        l2Address: chainId === l2.network.id ? address : undefined,
+        decimals,
+        listIds: existingToken?.listIds ?? new Set(),
+        lifiOnlyChainId: chainId,
+        priceUSD: existingToken?.priceUSD,
+      },
```
fionnachan: we should merge the `existingToken` with the new data
otherwise we lose the `logoURI`, `isL2Native`, etc props?

packages/arb-token-bridge-ui/src/util/mergeBridgeTokens.ts
fionnachan, line 1: this file is getting a little bit more complex now

please add a test file that tests this function directly

packages/arb-token-bridge-ui/src/hooks/useERC20L1Address.ts
```diff
@@ -12,20 +12,29 @@ import { getL1ERC20Address } from '../util/TokenUtils';
...
     async ([_eitherL1OrL2Address]) => {
-      const address =
-        (await getL1ERC20Address({
-          erc20L2Address: _eitherL1OrL2Address,
-          l2Provider,
-        })) ?? _eitherL1OrL2Address;
-      return address.toLowerCase();
+      const parentAddress = await getL1ERC20Address({
+        erc20L2Address: _eitherL1OrL2Address,
+        l2Provider,
```
fionnachan: should use `getProviderForChainId(_l2ChainId)` and not use the `l2Provider` directly

packages/arb-token-bridge-ui/src/components/TransferPanel/TokenSearch.tsx
fionnachan: it's super confusing that there are two on the source chain token search panel too
image

packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/utils.ts
```diff
@@ -54,6 +58,10 @@ export function isValidLifiTransfer({
     return false;
   }
 
+  if (allowsUnmatchedLifiTokens(sourceChainId)) {
```
fionnachan: why does it only consider source chain? what about destination chain?

fionnachan (approved):

PR #429, opened 2026-07-28
feat(bridge): disable CCTP v1

chrstph-dvx:
### Summary

image


packages/arb-token-bridge-ui/src/components/TransferPanel/Routes/CctpRoute.tsx
```diff
@@ -9,6 +9,17 @@ import { isNetwork } from '../../../util/networks';
 import { useRouteStore } from '../hooks/useRouteStore';
 import { Route } from './Route';
 
+export function CctpDeprecationWarning() {
+  return (
+    <div
+      role="alert"
+      className="rounded border border-orange-dark bg-orange-dark/50 p-3 text-sm text-white"
+    >
+      This route will be deprecated soon. Kindly claim your funds within the specific window.
```
fionnachan: Let's confirm with Milo what the specific window should be. We should state dates on the UI.

fionnachan (commented):
as discussed, banner should be attached to the CCTP v1 route only

let's also add the banner to tx history for users with unclaimed CCTP v1 txs

packages/arb-token-bridge-ui/src/components/common/SiteBanner.tsx
fionnachan, line 1: hm because we used connected wallet address here, note that searching an address won't show this banner

Image

`0x5c41796cbC77C9cB3f8b8F012c626c59cB39E03F`

fionnachan (approved):

dewanshparashar (approved):

PR #432, opened 2026-07-30
fix(bridge): re-enable Trust Wallet transfers

dewanshparashar:
Trust Wallet fixed the wrong-chain bug in mobile 26.30.6 and extension 2.91.4, so transfers are enabled again. Reverts the block from #400.

A dapp can't read the wallet's version, so Trust Wallet users see a short confirmation dialog before transferring instead.

## Changes

- Remove the block in `useTransferReadiness`
- Add `TrustWalletUpdateDialog`, shown on Move funds before any route
- Track `Trust Wallet Update Confirmation` for accept vs cancel rates

## Test plan

- Trust Wallet: dialog on Move funds; Confirm proceeds, Cancel aborts
- Other wallets: unchanged
- `pnpm lint` clean; tests match the master baseline

### DIFF
--- a/packages/arb-token-bridge-ui/src/util/AnalyticsUtils.ts
+++ b/packages/arb-token-bridge-ui/src/util/AnalyticsUtils.ts
@@ -42,6 +42,7 @@ type AnalyticsEventMap = {
     amount: number;
   };
   'Connect Wallet Click': { walletName: string };
+  'Trust Wallet Update Confirmation': { confirmed: boolean };
   'Fast Bridge Click': {
     bridge: FastBridgeName;
     tokenSymbol?: SpecialTokenSymbol.USDC;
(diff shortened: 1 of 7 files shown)


packages/arb-token-bridge-ui/src/util/AnalyticsUtils.ts
```diff
@@ -42,6 +42,7 @@ type AnalyticsEventMap = {
     amount: number;
   };
   'Connect Wallet Click': { walletName: string };
+  'Trust Wallet Update Confirmation': { confirmed: boolean };
```
chrstph-dvx, line 46: Nice idea to add analytics 👍

chrstph-dvx (approved):

PR #442, opened 2026-08-03
feat(bridge): resume LiFi routes from transaction history

chrstph-dvx:
## Stack

This is PR 2 of 3. It builds on #440, which adds multistep LI.FI route execution.

## Why

A multistep LI.FI transfer can complete its bridge step and still require another wallet action, such as a swap on the destination chain. The user can reject that later request, close the page, or reload before the route finishes.

Without this PR, transaction history can show the bridge transaction, but it cannot continue the remaining LI.FI steps. The user has no way to finish the route from the portal.

## What changes

This PR saves each LI.FI route update for transfers started in the portal. After a reload, transaction history reads the saved route, refreshes its status through LI.FI, and shows the full token path and the status of each step.

If LI.FI has stopped before all steps finish, transaction history shows a **Resume** action to the wallet that started the transfer. Resume uses the same chain switching and token approval flow as the original transfer.

The full LI.FI route is needed only while the transfer can continue. After the route completes or refunds, the cache removes transaction calldata and other execution data. It keeps the token amounts, tools, and step statuses that the history UI needs to render the completed transfer.

This PR also migrates existing versions 1 and 2 of the local LI.FI transaction cache to the new format.

## What to review

Please focus on these behaviors:

- An unfinished route survives a page reload with enough LI.FI data to continue the remaining steps.
- Completed and refunded entries no longer keep the full executable route, but transaction history still shows their token path and step statuses.
- LI.FI status responses update the current step without removing later steps that still need execution.
- **Resume** is hidden while LI.FI reports a step as `PENDING` or `STARTED`. Those states mean the SDK, wallet, or bridge may still be processing the step, so starting it again could create a duplicate request.
- An EIP-5792 wallet batch is not treated as stalled while its batch identifier may still resolve to a transaction hash. If it has no transaction link after 24 hours, the UI allows the sender to try again.
- Only the wallet that started the transfer can resume it.
- When the final resumed step reaches `DONE`, transaction history clears any failure status left by the earlier stopped attempt and marks the transfer as confirmed.

## Manual test

Arbitrum One WETH to Robinhood Chain SPCX with a low input amount will often return a route with multiple steps. Use an amount that is low but still returns a quote.

1. Select WETH on Arbitrum One as the source and SPCX on Robinhood Chain as the destination.
2. Enter a low amount. Confirm that the route summary shows at least two steps before you start the transfer.
3. Start the transfer and complete the first route step. Reject or stop at a later wallet request.
4. Open transaction history and reload the page.
5. Confirm that the transaction shows the token path, the completed step, the unfinished step, and a **Resume** action.
6. Connect a different wallet. Confirm that **Resume** is hidden.
7. Reconnect the original sender, click **Resume**, and accept the remaining requests.
8. Confirm that the transaction becomes complete and does not retain the failure status from the stopped attempt.
9. Reload transaction history. Confirm that the final token path and step statuses still render after the executable LI.FI route has been removed from the completed cache entry.

## Automated checks

```sh
pnpm --filter arb-token-bridge-ui exec vitest run src/components/TransactionHistory/TransactionDetailsContent.test.tsx src/components/TransactionHistory/TransactionsTableDetailsSteps.test.tsx src/components/TransactionHistory/TransactionsTableRowAction.test.tsx src/components/TransactionHistory/helpers.test.ts src/hooks/useLifiMergedTransactionCacheStore.test.ts src/hooks/__tests__/useTransactionHistory.test.ts
pnpm --filter arb-token-bridge-ui exec tsc --noEmit --pretty false
```


packages/arb-token-bridge-ui/src/components/TransactionHistory/helpers.ts
```diff
@@ -144,6 +159,56 @@ export function isTxPending(tx: MergedTransaction) {
...
+    return tx;
+  }
+
+  return {
+    ...tx,
+    status: WithdrawalStatus.CONFIRMED,
+    destinationStatus: isWithinLifiPendingPeriod(tx)
+      ? WithdrawalStatus.UNCONFIRMED
+      : WithdrawalStatus.CONFIRMED,
+  };
```
fionnachan: does it mean no matter what, the tx will be marked as `confirmed` after 24 hours?

packages/arb-token-bridge-ui/src/components/TransactionHistory/LifiTransactionDetails.tsx
fionnachan: Our normal tx details for pending txs have a "Get Help" button. Is that intentionally dropped here?

Image

fionnachan (changes requested):
testing this...
image
image

https://scan.li.fi/tx/0x179e32f8ec0f59bf1053bea47677254c7584c4315f8f6c8685c3dd9de491d5df
- if i re-enter the page (via new tab or refresh), the tx went to the `Settled` tab
- it is showing `success` on our tx history but also showing `resume`
- the external Li.Fi explorer link is only linking to the first step Wrapped eETH on Arbitrum -> ApeCoin on Robinhood Chain. need to change the final step's tx link.

fionnachan:
1. the tx pop ups immediately after i click the transfer button but we should have a dialog to inform users that they have to return to the page to resume the transaction within 24 hours or it would stop at one of the steps.
image

2. I rejected the transaction in my wallet but it appeared on the tx history
image

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -275,35 +294,42 @@ function mergeLifiTransaction({
     isWithdrawal: !isDepositMode,
     resolvedAt: apiTx.resolvedAt ?? localTx.resolvedAt,
     destinationTxId: apiTx.destinationTxId ?? localTx.destinationTxId,
-    durationMs: apiTx.durationMs || localTx.durationMs,
+    durationMs: Number.isFinite(apiSnapshot.durationMs)
+      ? apiSnapshot.durationMs
+      : localSnapshot.durationMs,
```
fionnachan: try `0`
`0 || 1` goes to 1
but `Number.isFinite(0) ? 0 : 1` goes to 0

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
```diff
@@ -1204,17 +1234,88 @@ export const useTransactionHistory = (
...
+
+      try {
+        const callsStatus = await getCallsStatus(wagmiConfig, { connector, id: batchId });
+        if (callsStatus.status === 'pending') {
+          return true;
+        }
+
+        const txHash = callsStatus.receipts?.at(-1)?.transactionHash;
+        if (!txHash) {
+          removeCachedTransaction(tx);
```
fionnachan: we should make sure that the searched address is the same as the connected address before we call `getCallsStatus`, otherwise they would end up at this row and we'll remove the cached tx

would be an issue if user switched wallet or is searching another address on tx history

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionsTableDetailsSteps.test.tsx
fionnachan: image
1. I rejected the transaction in my wallet but it appeared on the tx history
2. The "Resume" button shows up too early. The first few times I clicked it, no transactions showed in my wallet. I clicked it, then it says "Time left...", so I refreshed the page.

packages/arb-token-bridge-ui/src/hooks/useTransactionHistory.ts
chrstph-dvx: I'm not sure to understand? If durationMs is not a number, it should default to 0

fionnachan: if `apiSnapshot.durationMs` is 0, we return 0 with this new code

but the existing behaviour is to fallback to `localSnapshot.durationMs` if `apiSnapshot.durationMs` returns 0

fionnachan (commented):
the code looks fine to me

but in `packages/arb-token-bridge-ui/src/app/api/crosschain-transfers/lifi/transactions.ts` L176, the 0 seems like a fallback value, so I guess the question is whether we want to preserve `0` and treat it as the actual duration?

`durationMs: createdAt && resolvedAt ? Math.max(resolvedAt - createdAt, 0) : 0,`

fionnachan (approved):

PR #452, opened 2026-08-10
fix(bridge): load tokens on Token Search Panel first render

fionnachan:
### Summary

**The bug.** Opening the token search panel within the first seconds after load showed only the native currency row, and it never recovered. Reopening the dialog fixed it.

**What was happening.**

`useTokenListPriceUpdater.ts` was calling

```ts
const { mutate } = useTokenLists(childChain.id);
```

and because `useTokenLists` is called from multiple components, there was a race: another component starts the fetch first, and then this `mutate` runs while that fetch is still in flight.

SWR discards any fetch response that overlaps a `mutate` on the same key. Normally the mutation would schedule a fresh revalidation to recover, but `useTokenLists` uses `useSWRImmutable`, so it does not revalidate, and the fetched token lists are just lost.

So:

- If the user opens the token search panel (`TokenSearchUtils` calls `useTokenLists`) before that `mutate` finishes, the token list data is empty, and nothing ever refetches it.
- If the user opens the panel after the `mutate` has finished, the tokens load normally.

Reopening the dialog worked because it remounts `TokenSearch`, and a brand new subscriber runs its own fetch with no mutation in the way.

The `mutate` comes from the mount effect in `useTokenListPriceUpdater`, which lives in `MainContent`, so it fires on every bridge page load.

**The fix.** Return early from `refreshLifiTokenList` until `useTokenLists` has data, so the price refresh can never overlap the initial fetch.

Two notes for reading the diff:

- The guard is a `hasTokenLists` **boolean**, not `tokenLists` itself. `mutate` writes a new array, so depending on the array would change the callback identity on every refresh and the effect would re-trigger it in a loop.
- "Has data" is a sufficient guard only because `useTokenLists` never revalidates once it has data. That assumption is called out in a comment at the check.

**Not the fix:** changing what the `mutate` callback returns. The response is discarded regardless of what it returns, so the mutation simply must not be in flight during that window.

**Alternative considered:** giving the price refresh its own SWR key. It works, but `refreshInterval` then lands on a hook read from 18 call sites (two of them per-row inside the virtualized token list), so every call site schedules its own timer, and dropping `useSWRImmutable` silently changes the revalidate-on-mount/focus/reconnect behaviour. Guarding the existing `mutate` keeps polling on the single `useTokenListPriceUpdater` instance, as originally designed.

**Out of scope:** if the load-time LiFi request fails there is no list entry for a later refresh to merge into, so refreshed tokens are dropped from the token lists (they still reach `bridgeTokens` via `addTokensFromList`). Pre-existing, unchanged here.

### Steps to test

**Manual repro:** hard-reload `/bridge?sourceChain=ethereum&destinationChain=arbitrum-one`, then click the token selector within the first few seconds, while the token list request is still in flight.

- Before: 1 row (`ETH`), never recovers.
- After: full list on the first open (113 rows in Chromium, matching what the second open gives).

Also checked on `ethereum -> apechain`: 5 rows immediately (`APE, ETH, WETH, USDT, USDC`), price poll fires once at +32.8s, so the refresh itself still works.

**Automated:**

```bash
# Node 22 (see .nvmrc). On Node 26, happy-dom's localStorage is shadowed and unrelated tests fail.
cd packages/arb-token-bridge-ui
pnpm vitest run src/hooks/__tests__/useTokenListPriceUpdater.test.tsx
```

The new test mounts the price updater alongside `useTokenLists` with a delayed fetch so the two overlap. Both cases pass here and fail when the hook is reverted to `master`.

`pnpm lint` and `pnpm prettier:check` are clean. Full `pnpm test:ci` has one failure, `TransferPanel.swap.integration.test.tsx > 'base' -> 'robinhood-chain'`, which fails identically on unmodified `master`.

### References

- Introduced by 69dec824a `feat(bridge): add LiFi swap (#150)`
- The SWR guard that discards the response: `swr@2.3.3`, `dist/index/index.mjs:382-396`

Closes FS-2554

### DIFF
--- a/packages/arb-token-bridge-ui/src/hooks/__tests__/useTokenListPriceUpdater.test.tsx
+++ b/packages/arb-token-bridge-ui/src/hooks/__tests__/useTokenListPriceUpdater.test.tsx
@@ -10,12 +10,24 @@ import { useNetworksRelationship } from '../useNetworksRelationship';
 import { useTokenListPriceUpdater } from '../useTokenListPriceUpdater';
 import { useTokenLists } from '../useTokenLists';
 
-const { FETCH_DELAY_MS, buildTokenList } = vi.hoisted(() => ({
-  // Long enough that the initial useTokenLists fetch is still in flight when the price
-  // updater mounts, which is the window the regression lived in.
-  FETCH_DELAY_MS: 50,
-  buildTokenList: (symbol: string) => ({
-    name: 'Test list',
+const { FETCH_DELAY_MS, MOCK_BRIDGE_TOKEN_LISTS, MOCK_TOKEN_LISTS } = vi.hoisted(() => {
+  const arbitrumTokenList = {
+    name: 'Mock Arbitrum Token list',
+    timestamp: '2026-01-01T00:00:00.000Z',
+    version: { major: 1, minor: 0, patch: 0 },
+    tokens: [
+      {
+        chainId: 42161,
+        address: '0x912ce59144191c1204e64559fe8253a0e49e6548',
+        name: 'Arbitrum',
+        symbol: 'ARB',
+        decimals: 18,
+      },
+    ],
+  };
+
+  const lifiTokenList = {
+    name: 'Mock LiFi list',
     timestamp: '2026-01-01T00:00:00.000Z',
     version: { major: 1, minor: 0, patch: 0 },
     tokens: [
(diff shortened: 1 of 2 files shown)


packages/arb-token-bridge-ui/src/hooks/useTokenListPriceUpdater.ts
```diff
@@ -21,14 +21,34 @@ export function useTokenListPriceUpdater({
...
+    /**
+     * Never mutate this key while its own fetch is in flight. SWR discards any response that
+     * overlaps a mutation on the same key, and `useTokenLists` is `useSWRImmutable`, so there is no
+     * revalidation left to recover the discarded token lists with: every consumer would be stuck
+     * with no tokens until a new subscriber mounted.
+     *
+     * Data being present is a sufficient check only because `useTokenLists` never revalidates once
+     * it has data. Revisit this if that changes.
+     */
+    if (!hasTokenLists) {
```
chrstph-dvx, line 48: Should not we wait on loading instead? If loading is done and tokenLists is undefined, we can still mutate no?

packages/arb-token-bridge-ui/src/hooks/__tests__/useTokenListPriceUpdater.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+const childChain = { id: ChainId.ArbitrumOne };
+
+function wrapper({ children }: PropsWithChildren) {
+  // Fresh cache per render, so one test cannot satisfy another from cache.
+  return (
+    <SWRConfig value={{ provider: () => new Map(), dedupingInterval: 0 }}>{children}</SWRConfig>
+  );
+}
+
+describe.sequential('useTokenListPriceUpdater', () => {
```
chrstph-dvx, line 75: Do we need sequential if we have one wrapper per test?

packages/arb-token-bridge-ui/src/hooks/__tests__/useTokenListPriceUpdater.test.tsx
```diff
@@ -0,0 +1,132 @@
...
+      },
+      { wrapper },
+    );
+
+    // The updater mutates the useTokenLists key. SWR discards any response that overlaps a
+    // mutation on the same key, and useSWRImmutable has no revalidation left to recover
+    // with, so refreshing during the initial fetch used to leave data undefined forever.
+    await waitFor(
+      () => {
+        expect(result.current.data?.length).toBeGreaterThan(0);
```
chrstph-dvx, line 105: Can we assert for the exact result? Same on L125

chrstph-dvx (approved):

PR #464, opened 2026-08-14
feat(bridge): add search by hash in tx history

dewanshparashar:
Adds a dropdown to the tx history search bar to search by tx hash, next to the existing wallet address search.

A hash search looks up the receipt directly instead of scanning event logs from genesis. On chains with aggressive RPC rate limits (like Edu Chain) this loads the tx in a few seconds instead of hours, so users can find and claim their withdrawal right away.

Notes:
- Works for canonical deposits and withdrawals, plus CCTP, OFT and LiFi transfers
- The hash must be the source chain tx that initiated the transfer
- Only the chains selected in the chain filter are searched, to keep RPC usage low. The not-found message points at the chain filter
- While a hash search is active, the slow full history fetch does not run. useTransactionHistory keeps its original signature and reads the searched hash from the search store
- Searching a tx that does not belong to the connected wallet shows a warning banner in place of the claim reminder
- The search bar is restyled to the design spec: mode dropdown with divider, no search button, searches on enter or on clicking out of the field, red border on invalid input

Known limitation: transfers initiated by a smart contract wallet may not resolve for deposits and CCTP/LiFi, because the receipt sender is the relayer rather than the wallet. Withdrawals are unaffected.

Testing: unit tests for the hash search fetcher, receipt parsing and the search store. Verified locally with Edu Chain withdrawals, found in a few seconds. Lint and test:ci pass.


packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistorySearchBar.tsx
```diff
@@ -10,74 +11,205 @@ import { Tooltip } from '@/app/components/common/Tooltip';
...
-  }, [address, connectedAddress, setSanitizedAddress, setSearchError]);
-
-  const searchTxForAddress = useCallback(() => {
     if (address === '') {
-      return;
+      setSanitizedTxHash(undefined);
+      if (connectedAddress) {
+        setSanitizedAddress(connectedAddress);
+        setSearchError(undefined);
+      }
```
chrstph-dvx: `address` is coming from `useTransactionHistoryAddressStore` already, if address changes, why do we need to call `setSanitizedAddress` here?

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistorySearchBar.tsx
```diff
@@ -10,74 +11,205 @@ import { Tooltip } from '@/app/components/common/Tooltip';
...
+    },
+    [address, searchTx, setSearchMode],
+  );
 
-    setSanitizedAddress(address);
-    setSearchError(undefined);
-  }, [address, setSanitizedAddress, setSearchError, isTestnetMode, connectedAddress]);
+  // Auto-search when the user clicks out of the field, but only with valid
+  // input: a blur from switching the search mode must not flash an error.
+  const searchTxIfValid = useCallback(() => {
```
chrstph-dvx: If you type an invalid tx hash, and click outside of the input. This will keep the invalid txHash, but it will display result for the previous valid tx hash

chrstph-dvx (commented):
Address and tx hash have different format, can we remove the dropdown for user, and detect whether it's an address or a transaction?

chrstph-dvx:
If I search for a settled transaction, by default it shows me "No pending transactions", it would be nice, if we found a settled transaction, to switch tab for the user

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistorySearchBar.tsx
dewanshparashar: That line is existing behavior from master: clearing the input falls back to the connected wallet's history. My edit had wrapped it inside a new if block which made it show as new in the diff. Restructured in 84f2584 so the diff only shows the new tx hash reset line.

packages/arb-token-bridge-ui/src/components/TransactionHistory/TransactionHistorySearchBar.tsx
dewanshparashar: Good catch, fixed in 84f2584. Clicking out of the field now validates like a submit, so an invalid hash shows the error instead of keeping the previous result.

dewanshparashar:
Done in 6ea26e6. When a hash search finds only a settled tx, the settled tab opens for the user. It runs once per searched hash, so switching tabs manually afterwards still sticks.

dewanshparashar:
It already picks the tab from the transaction status: the pending and settled groups come from the same status checks the tabs render with (pending, claimable and failed go to the pending tab, completed and expired to the settled tab). The count check just read badly, renamed it in abd4955 to make the intent clear.

chrstph-dvx (approved):

PR #471, opened 2026-08-21
fix(bridge): select OFT transfer log by emitter

dewanshparashar:
USDT0 transaction history read `receipt.logs[0]` and assumed it was the token's `Transfer` event. EIP-7708 adds protocol ETH transfer logs ahead of the EVM logs, so that position will point at the wrong entry once Glamsterdam ships. The protocol log has the same shape as an ERC20 `Transfer`, so it decodes fine and silently yields the ETH fee as the transfer amount.

Now matches on the token address and the `Transfer` topic instead of the position.

No behaviour change today. On current receipts the returned values are identical, verified by running the integration test against both the old and new code.

Tests:

- unit test with a protocol ETH transfer log prepended to the receipt (fails on the old code with `2500000000.0` instead of `1.5`)
- integration test against a real mainnet USDT0 send

FS-2552

### DIFF
--- a/packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.integration.test.ts
+++ b/packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.integration.test.ts
@@ -3,7 +3,6 @@ import { describe, expect, it } from 'vitest';
 
 import { LayerZeroTransaction } from '../state/app/state';
 import { ChainId } from '../types/ChainId';
-import { addressesEqual } from '../util/AddressUtils';
 import { CommonAddress } from '../util/CommonAddressUtils';
 import { AssetType } from './arbTokenBridge.types';
 import { updateAdditionalLayerZeroData } from './useOftTransactionHistory';
@@ -13,6 +12,7 @@ import { updateAdditionalLayerZeroData } from './useOftTransactionHistory';
 // is what makes EIP-7708 prepend a protocol Transfer log to this receipt after Glamsterdam.
 const USDT0_SEND_TX = '0xb7ccb326d0b383eea44343a60e6f0563c7a7e0a5d3b96696ea39bb7cf76fe3ad';
 const USDT0_SEND_BLOCK = 25799541;
+const USDT0_SEND_DESTINATION = '0x36d16e71d630ba8c5bfb7ac8a459aebdac1276aa';
 const USDT_AMOUNT = '2500.0';
 // what the amount would read as if the ETH fee were decoded as USDT instead
 const ETH_FEE_AS_USDT = '37964723.338644';
@@ -37,7 +37,7 @@ const transaction = {
   sourceChainId: ChainId.Ethereum,
   destinationChainId: ChainId.ArbitrumOne,
   destinationTxHash: null,
-} as LayerZeroTransaction;
+} satisfies LayerZeroTransaction;
 
 describe('updateAdditionalLayerZeroData against a real USDT0 transfer', () => {
   it('resolves the transferred token amount from the live receipt', async () => {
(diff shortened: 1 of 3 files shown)


packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.integration.test.ts
```diff
@@ -0,0 +1,55 @@
...
+  uniqueId: null,
+  isWithdrawal: false,
+  blockNum: null,
+  tokenAddress: CommonAddress.Ethereum.USDT,
+  childChainId: ChainId.ArbitrumOne,
+  parentChainId: ChainId.Ethereum,
+  sourceChainId: ChainId.Ethereum,
+  destinationChainId: ChainId.ArbitrumOne,
+  destinationTxHash: null,
+} as LayerZeroTransaction;
```
chrstph-dvx: NIT: I'd use satisfies here instead

packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.integration.test.ts
```diff
@@ -0,0 +1,55 @@
...
+
+describe('updateAdditionalLayerZeroData against a real USDT0 transfer', () => {
+  it('resolves the transferred token amount from the live receipt', async () => {
+    const result = await updateAdditionalLayerZeroData(transaction);
+
+    expect(result.value).toBe(USDT_AMOUNT);
+    expect(result.value).not.toBe(ETH_FEE_AS_USDT);
+    expect(result.asset).toBe('USDT');
+    expect(addressesEqual(result.tokenAddress ?? undefined, CommonAddress.Ethereum.USDT)).toBe(
```
chrstph-dvx: Test should be deterministic, it should either be result.tokenAddress or undefined IMO

packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.integration.test.ts
```diff
@@ -0,0 +1,55 @@
...
+      true,
+    );
+    expect(result.blockNum).toBe(USDT0_SEND_BLOCK);
+    expect(utils.isAddress(result.destination ?? '')).toBe(true);
```
chrstph-dvx: same here

packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.test.ts
```diff
@@ -0,0 +1,129 @@
...
+    uniqueId: null,
+    isWithdrawal: false,
+    blockNum: null,
+    tokenAddress,
+    childChainId: ChainId.ArbitrumOne,
+    parentChainId: ChainId.Ethereum,
+    sourceChainId: ChainId.Ethereum,
+    destinationChainId: ChainId.ArbitrumOne,
+    destinationTxHash: null,
+  } as LayerZeroTransaction;
```
chrstph-dvx: NIT: I'd use satisfies here

packages/arb-token-bridge-ui/src/hooks/useOftTransactionHistory.test.ts
```diff
@@ -0,0 +1,129 @@
...
+
+describe('updateAdditionalLayerZeroData', () => {
+  it('reads the token transfer when a protocol ETH transfer log is emitted first', async () => {
+    const tx = oftTransaction('0xaa', [protocolEthTransferLog, usdtTransferLog]);
+
+    const result = await updateAdditionalLayerZeroData(tx);
+
+    expect(result.value).toBe('1.5');
+    expect(result.asset).toBe('USDT');
+    expect(addressesEqual(result.tokenAddress ?? undefined, CommonAddress.Ethereum.USDT)).toBe(
```
chrstph-dvx: Same here: I'd set one of the value, we should know whether we have a tokenAddress or not

chrstph-dvx (approved):

PR #475, opened 2026-08-27
refactor(bridge): serve native bridge history from the indexer

douglance:
Stacked on #469. The indexer has backfilled native bridge history for **Arbitrum One (42161)** and **Arbitrum Sepolia (421614)**, so this drops the `l1-*` / `l2-*` subgraphs behind them. **Arbitrum Nova** is not backfilled and keeps its subgraphs, so this is a narrowing rather than the deletion — see *Endgame* below.

### The flip is config; this makes it load-bearing

Adding `42161,421614` to `NEXT_PUBLIC_INDEXER_CHILD_CHAIN_IDS` already routes both chains to the indexer today, with the subgraph sitting behind them as an accidental second source that can silently disagree. Removing the subgraph is what makes the config the single answer.

**Blocking prerequisite:** `NEXT_PUBLIC_INDEXER_CHILD_CHAIN_IDS` must list `42161,421614` in Vercel prod + preview **and** as a GitHub repo secret before this merges. It is `NEXT_PUBLIC_`, so it is baked at build time; it is now passed to `build.yml`, `test.yml`, and `e2e-tests.yml`, where it was previously absent. `INDEXER_API_URL_BY_CHAIN` already needs entries for both chains (they are CCTP chains, per #469), so no new URLs are required.

### Fail loudly — generalizing @dewanshparashar's #469 review

`proxyToIndexer` answers a failure with `{ data: [], message: 'Indexer unavailable' }` at 502, and all three history fetchers read `.data` without checking `response.ok`. The error body's own empty array was being consumed as a successful empty history — the same trap as `fetchCCTP`, and removing the subgraph is what makes it load-bearing here.

For withdrawals it also costs the recovery path: `fetchWithdrawals` widens its event-log scan to the full range only if the fetch *throws*, so returning `[]` skipped it. Deposits have no event-log path (`// TODO : Add event logs as well`), so there the gain is a logged error rather than an invisible one — stating that plainly rather than overselling it.

### Mainnet is covered by CI, unusually

`test.yml` runs `pnpm test:unit` behind a real server (`start-server-and-test`), and `src/util/{deposits,withdrawals}/__tests__/fetch*FromSubgraph.test.ts` assert exact result counts and real transaction hashes for Arbitrum One block ranges. After this change those assertions read from the indexer. They are the strongest pre-merge signal that the backfill matches the subgraph — and if they go red, that is data disagreement, not a flaky test. They fail locally without a dev server, as they did before this PR.

Still worth a manual pass on the preview deployment for pagination (`Load More`) and the settled/claimable split, which no test covers.

### Endgame

Once Nova is backfilled, a follow-up deletes the `subgraphs` map, `createSubgraphClient` / `createSelfHostedSubgraphClient` / `createTheGraphNetworkClient` / `SourcedClient` / the `Subgraph` union, the gql bodies in all three route files, `hasBridgeSubgraph`, the block-number subgraph branch, and both subgraph API keys — roughly 500 lines.

Deliberately not here: renaming the `*FromSubgraph` surface (large churn, and `WithdrawalFromSubgraph.source: 'subgraph'` is a shape discriminant read by `useTransactionHistory`, not a provenance label), and server-side Sentry for the route catches, already promised as its own PR.


douglance:
Closing — opened ahead of time. Reopening restores the description and commits.

